PR #166, opened 2014-12-07
More informative ProcessRegistrationException.

mboes:
Now includes the identifier of the process that owns the name, if any.
The exception used to only mention the name. However, that made it
impossible to implement many algorithms, such as "return registered
process or spawn new one" in a race free way. If you catch
ProcessRegistrationException after attempting `register`, doing
`whereis` after that to find out who owns the name is racy.

This is a BREAKING CHANGE, albeit to a peripheral part of the current
API, so the next release of distributed-process that includes this
change will have to bump the major version number according to the PVP.

Needs companion patch to distributed-process-tests to pass CI successfully.


hyperthunk:
Soon, soon. Just working through some more minor issues and then I'll start updating everything to use this new API.

hyperthunk:
This is still on hold. We should have a release out this month (I'm working on it!) and I'll start refactoring for this change immediately afterwards.

mboes:
@facundominguez now that we've pushed a release, could we get this merged?

PR #169, opened 2015-01-01
DP-02: simple approach for maintaining queue size.

qnikst:
This commit introduces basic approach to maintain size of the
queue by keeping TVar Int in the state, then each writer will
increment this state, and deque will decrement it if and only
if it outputs value.

### DIFF
--- a/src/Control/Distributed/Process/Internal/CQueue.hs
+++ b/src/Control/Distributed/Process/Internal/CQueue.hs
@@ -10,6 +10,7 @@ module Control.Distributed.Process.Internal.CQueue
   , enqueueSTM
   , dequeue
   , mkWeakCQueue
+  , queueSize 
   ) where
 
 import Prelude hiding (length, reverse)
@@ -17,10 +18,14 @@ import Control.Concurrent.STM
   ( atomically
   , STM
   , TChan
+  , TVar
+  , modifyTVar'
   , tryReadTChan
   , newTChan
+  , newTVarIO
   , writeTChan
   , readTChan
+  , readTVarIO
   , orElse
   , retry
   )
@@ -38,6 +43,7 @@ import Control.Distributed.Process.Internal.StrictList
   , append
   )
 import Data.Maybe (fromJust)
+import Data.Foldable (traverse_)
 import GHC.MVar (MVar(MVar))
 import GHC.IO (IO(IO))
 import GHC.Prim (mkWeak#)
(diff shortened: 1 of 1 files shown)


hyperthunk:
I'm not sure why we need that `traverse_` here - why not just update the field directly? I may be betraying a lack of haskell-mindedness there..

Also, this may have been a problem with the existing code and nothing introduced here, but... Aren't we potentially going to have a problem here if we dequeue from `incoming` but then an asynchronous exception blows us up before we decrement the `size` field? In fact, don't we need to do something in `grabNew` as we're pulling new messages from the `incoming` into the `arrived` field too?

@edsko, @mboes, @dcoutts, @facundominguez - any comments about this? Having the mailbox size is a *very* useful thing for a variety of algorithms btw, so I'm not asking about the usefulness of the PR - though we can bike shed if you really want to ;-) - but are we on the right track? Should we _mask_ the whole operation including the size field updates, or is that overkill?

qnikst:
I think the correct solution depends on task about safe `timeout` handling (can't find email about it) and could be solved together. Otherwise this code is not exception safe. It seems that we may want to mask all the operation and release mask in relevant places. However I'd like to gather all feedback before.

hyperthunk:
Thanks @qnikst - I think you're absolutely right about that. See https://cloud-haskell.atlassian.net/browse/DP-98 for additional background regarding the potential problems that using `System.Timeout` introduces.

edsko:
I'm not sure what you're asking me here. But if you are `TVar`s, i.e. STM, then you should not need to have to worry about masking. Instead make sure that things you want to run as a single unit are a single transaction and the STM infrastructure will take care of the rest. But you probably know this already so perhaps I misunderstand what you're asking.

mboes:
I'm concerned that decrementing is not done atomically with popping an item off the queue (incrementing OTOH looks ok). I agree with @edsko - there is no need to worry about exceptions, async or otherwise. But that can only hold true if incrementing and decrementing the size is done atomically with the rest. Providing atomic updates to concurrent state without too much headache is exactly what STM is about. Let's make use of it accordingly!

qnikst:
Then there is a question we have a StrictList inside MVar, so we have a bit of IO here, that is not composable with STM. Is it for purpose, can we use TVar instead? Is we can then dequeue can be an STM operation. I see 2 possible reasons:
1. We want to have timeout, but its possible  to have it with STM (having `registerDelay`)
2. We want to have ordering i.e. all threads that call dequeue should return results in order, its not possible to implement in one STM transaction.
3. We don't want to wake up all threads that is 'locked' on TVar (not sure it matters).

If latter matter then we don't have a way to use STM only and should care about exceptions.

I gave used TVar instead on IORef as it plays nicely with enqueue that is implemented as a single STM transaction.

hyperthunk:
See my latest comment in https://github.com/haskell-distributed/distributed-process/pull/170 - I think we want to move to `TVar` there as well, but there's a remaining question about strictness properties that needs answering.

hyperthunk:
> 1. We want to have timeout, but its possible to have it with STM (having registerDelay)
2. We want to have ordering i.e. all threads that call dequeue should return results in order, its not possible to implement in one STM transaction.
3. We don't want to wake up all threads that is 'locked' on TVar (not sure it matters).

1. I can't see what use `registerDelay` would be here TBH. I think we still have to use `timeout` but the whole STM transaction should fail if we time out won't it?

2. Only 1 thread ever calls dequeue at a time.

3. I think we want `TMVar` here anyway.

> If latter matter then we don't have a way to use STM only and should care about exceptions.

Even that isn't a solution because we'd need `uninterruptableMask` and that introduces a whole different set of problems according to @edsko's comment in https://github.com/haskell-distributed/distributed-process/pull/170. 

> I gave used TVar instead on IORef as it plays nicely with enqueue that is implemented as a single STM transaction.

Agree it makes sense.

hyperthunk:
So, looking at the conversation we've had over on https://github.com/haskell-distributed/distributed-process/pull/170, it seems that we're _ok_ to continue using an `MVar` for that patch, but here we want to keep the {en,de}queue and size counter {in,de}crement atomic so we really want to use *only* STM operations if possible. We need to make a call on these two patches as they do affect each other somewhat.

qnikst:
This patch will require to move mask_ on the top, I.e. `dequeue = mask_ $ ...`. I can try to write STM only solution, but we will have to carefully benchmark it.

hyperthunk:
@qnikst - ok let's get #170 merged and look at an STM based solution.

qnikst:
I'm closing this pull request for now.

hyperthunk:
@qnikst why?

qnikst:
There are 2 reasons for this: 
1. Depends on #170 and have to be rewritten if #170 will be merged (small fixes but anyway)
2. If we will go with STM based solution with patch also  should be rewritten

So at this moment this a PR that can't be applied. I do not remove branch so anybody (or me) can modify it and reopen this PR once #170 will be merged or this patch will be done in terms of STM. I hope to do it on a weekend.

PR #170, opened 2015-01-05
Prevent timeouts from losing messages in CQueue.

facundominguez:
A fix for https://cloud-haskell.atlassian.net/browse/DP-98.

### DIFF
--- a/src/Control/Distributed/Process/Internal/CQueue.hs
+++ b/src/Control/Distributed/Process/Internal/CQueue.hs
@@ -101,7 +101,7 @@ dequeue :: forall m a.
         -> BlockSpec         -- ^ Blocking behaviour
         -> [MatchOn m a]     -- ^ List of matches
         -> IO (Maybe a)      -- ^ 'Nothing' only on timeout
-dequeue (CQueue arrived incoming) blockSpec matchons =
+dequeue (CQueue arrived incoming) blockSpec matchons = mask_ $
   case blockSpec of
     Timeout n -> timeout n $ fmap fromJust run
     _other    ->
@@ -115,7 +115,7 @@ dequeue (CQueue arrived incoming) blockSpec matchons =
   where
     chunks = chunkMatches matchons
 
-    run = mask_ $ do
+    run = do
            arr <- takeMVar arrived
            let grabNew xs = do
                  r <- atomically $ tryReadTChan incoming


hyperthunk:
I don't believe so. Can we add one as part of this changeset?

hyperthunk:
@edsko, @facundominguez, @mboes - do we need to worry about the overhead of masking every dequeue operation? I'm not saying I've got a viable alternative, just wondering???

facundominguez:
I think I need some inspiration. How would you test that the mask is preventing the timeout from dropping messages?

edsko:
@hyperthunk I have not really been following this development at all so I don't really know what's going on. Masking doesn't have much overhead, it just adds a single word onto the runtime stack, and then a check for delayed exceptions when leaving the scope of the mask. However, a quick glance at the commit shows that you have interruptible operations inside your `mask_` (`takeMVar`), for instance. Note that these are _still_ interruptible, even when using `mask_`. If you don't want this, you will need to use `uninterruptibleMask_` but this comes with big problems of its own. Getting such design decisions completely correct is a non-trivial task! (If using STM is a viable alternative, I think you should consider it.)

hyperthunk:
Thanks @edsko, that's exactly the sort of reason I wanted some peer review. So it seems using `mask_` isn't enough in this situation, since we can still theoretically loose messages that way. In https://github.com/haskell-distributed/distributed-process/pull/169 there's a desire to store the `StrictList` inside `CQueue` in a `TVar` instead of a `StrictMVar` since that way we can avoid the need for masking in that situation. Can we kill both birds with one stone then by changing the `arrived` field to be a `TVar` instead of a `StrictMVar`? And if we do, what are the strictness properties of `TVar` - do we need to write a `StrictTVar` too?

hyperthunk:
@mboes, @facundominguez - anything to add here?

edsko:
Strictness properties of `TVars` and `MVars` are the same.

hyperthunk:
@edsko - ok so we'll need to write a `ScrictTVar` wrapper then. Not sure I'm the right person to do that - @mboes, @facundominguez either of you fancy it? Fixing that will sort us out for https://github.com/haskell-distributed/distributed-process/pull/169 as well.

qnikst:
Just to remind are we OK with loosing fairness property that will not hold with STM?

hyperthunk:
Erm, sheesh no actually we're not. Hmn.

hyperthunk:
> Erm, sheesh no actually we're not. Hmn.

Wait, let's step back a moment. You're talking about fairness and that *only* matters when we're talking about multiple writers, because there's only every one thread acting as a reader. Well, this is not an issue in the current implementation. All sending is done via the node controller, so there's only ever one thread that's writing messages to the `incoming` field. The same is true for reading, because each `Process` has its own mailbox and only the current process can read from its own `CQueue` so we shouldn't have to worry about scheduling/fairness here at all. The only thing we _must_ consider is that we cannot change the `localSend` implementation to bypass the node controller and write directly to the destination processes mailbox, because that *would* introduce potential problems. That was only suggested as an optimisation and hasn't been implemented, so I think we're fine.

qnikst:
I was thinking about fairness on reading, if there are multiple readers we can be in a situation where one reader will have a preference (e.g. it took less time to read event as they for example in the beginning of the queue then all others will retry), I think I can write an example.

facundominguez:
> So it seems using mask_ isn't enough in this situation, since we can still theoretically loose messages that way.

I don't follow here. How can a message be lost exactly?

If this solution turns out to not introduce new issues, it could be a simple interim solution while someone else works out a better implementation. I spent some time reviewing CQueue.hs a while ago and it looks safe (though it is not as obvious as I would wish).

edsko:
`putMVar` is only interruptible when the `MVar` is not empty, so in a typical `takeMVar ... putMVar` scenario you are okay (provided that all concurrent threads do the same `takeMVar ... putMVar`. `takeMVar` is a different story though. You are right that `takeMVar` is interruptible only when the `MVar` is empty, but in typical scenarios this is not something you cannot guarantee not to be the case. 

As regards `uninterruptibleMask_`: the main problem is that you are now blocking _everything_ and you have to be very careful. Generally speaking `uninterruptibleMask_` should not be used for things that can take an arbitrary amount of time (like using `takeMVar` to wait on a lock). For example, check out my blog post http://www.well-typed.com/blog/97/, the corresponding Reddit page http://www.reddit.com/r/haskell/comments/2ety9f/new_blog_post_dealing_with_asynchronous/, and a proposal (now accepted, I think) to use `uninterruptibleMask_` for cleanup in `bracket` on the `libraries` mainling list https://www.haskell.org/pipermail/libraries/2014-September/023675.html.

hyperthunk:
Right, thanks very much for the clarification and additional info @edsko - that's been really helpful for me. @facundominguez - because we only have one reader I think we should get rid of the `MVar` for `arrived`. Because we have three threads contending for `incoming` however, it follows that either one of the two node controller event loop threads might hold the lock when the reading process tries to `takeMVar` but that's not going to cause any issues. Now if the mailbox owner (process) has acquired the lock it is not going to block when it evaluates `putMVar` later on and if either or the NC threads is blocking and an asynchronous exception arrives, well we have bigger problems if _that_ code isn't exception safe (which I'm pretty sure it is anyway).

So it now *sounds* to me like we're ok with what's being proposed in this patch. STM might be cleaner but we can revisit that. Does this decision affect the outcome of https://github.com/haskell-distributed/distributed-process/pull/169 @qnikst / @mboes? I don't think so but worth considering before I merge.

qnikst:
As we should care about correctness more than about functionality this patch should be merged first, then I'll revisit #169. It will introduce a small changes but I don't want to mess discussion before this PR will be handled.

hyperthunk:
Ok @qnikst - thanks for that. I'll concentrate on getting this merged first. @facundominguez - we still need to think about whether a regression test is plausible.

hyperthunk:
Ok @facundominguez I'm going to merge without a regression test because I want to get this in and start thinking about another release in a week or two. Should be done over this weekend...

facundominguez:
Alright. I'm afraid I hadn't the time to explore with ways to expose this bug.

For the record, a possible test is to build CQueue.hs with a custom-made version of `timeout`, which delivers an asynchronous exception every time the action completes.

In any case, if this is going to be reorganized to use STM, it wouldn't worth spending much effort in testing it.

hyperthunk:
bq. In any case, if this is going to be reorganized to use STM, it wouldn't worth spending much effort in testing it.

Ok agreed. I'm going to merge this shortly.

edsko:
So I didn't look at this very carefully except to answer Tim's question above, but nice catch @facundominguez .

PR #171, opened 2015-01-11
Introduce callLocal primitive.

qnikst:
'callLocal' provides a way to create processes that are isolated
by their ProcessId, so one can send messages to that processes
iff that process announced his id, and all messages that were sent
to that process after it dies will be silently discarded.

Original code author: Miëtek Bak <mietek.bak@tweag.io>

### DIFF
--- a/src/Control/Distributed/Process.hs
+++ b/src/Control/Distributed/Process.hs
@@ -149,6 +149,7 @@ module Control.Distributed.Process
     -- * Local versions of 'spawn'
   , spawnLocal
   , spawnChannelLocal
+  , callLocal
     -- * Reconnecting
   , reconnect
   , reconnectPort
@@ -302,6 +303,7 @@ import Control.Distributed.Process.Internal.Spawn
   , spawnSupervised
   , call
   )
+import Control.Exception (SomeException, throwIO)
 
 -- INTERNAL NOTES
 --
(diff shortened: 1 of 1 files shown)


qnikst:
This primitive is very usefull to have for a number of use cases where it's ok to loose some messages after the work was done, but messing results with previous process invocation may lead to a problems. Other approaches like tags could work but are more heavyweight for those usecases.

src/Control/Distributed/Process.hs
```diff
@@ -392,3 +392,17 @@ spawnChannelLocal proc = do
...
+
+--- Local version of 'call'. Running a process in this way isolates it from
+--- messages sent to the caller process, and also allows silently dropping late
+--- or duplicate messages sent to the isolated process after it exits.
+--- Silently dropping messages may not always be the best approach.
+callLocal ::
+     Process a  -- ^ Process to run
+  -> Process a  -- ^ Value returned
+callLocal proc = do
+    mv <- liftIO newEmptyMVar
```
hyperthunk: What's this `MVar` being used for?

hyperthunk:
Sorry @qnikst but this patch seems broken. The implementation doesn't do anything (?) so I'm assuming we're looking at a borked or incomplete merge. Let me know when it's ready though! :)

qnikst:
You are right, fixed.

qnikst:
The patch was updated, can you review it once again? Thanks.

src/Control/Distributed/Process.hs
```diff
@@ -392,3 +392,20 @@ spawnChannelLocal proc = do
       liftIO $ putMVar mvar sport
       proc rport
     takeMVar mvar
+
+--- Local version of 'call'. Running a process in this way isolates it from
```
mboes: `---` as a comment delimiter? Lack of `-- |` to open it?

qnikst: Thanks, fixed in 4e92716

qnikst:
Thanks, @hyperthunk and @mboes.

I've recreated a patch, also codestyle was fixed in order to match style in the file.

hyperthunk:
Much better, thanks. I want to get CI running/fixed before merging so this will probably get done either over the weekend or on Monday.

PR #172, opened 2015-01-15
Fix travis build

qnikst:
Remove upper bound restriction on random package in order
to fix travis build.


qnikst:
This commit will allow Travis to check Pull request again!

PR #173, opened 2015-01-15
Implement unreliable send (usend).

facundominguez:
Unlike `send`, the function `usend` is insensitive to `reconnect`. It will try to send the message regardless of the history of connection failures between the nodes.

It saves the tediousness of dealing with 'monitor' and 'reconnect' calls when reliability is not required.

The implementation mimics `nsendRemote`. The `msg` and the `pid` go together to the target node, and the target node dispatches the message to the target process.

When working in a network with failures, distributed algorithms sometimes don't have reliability or ordering requirements. Using `usend` saves many brain cycles that would be spent in figuring out what to `monitor` and when to call `reconnect`.

If this implementation is liked, we would add some tests for it in d-p-tests.

Maybe `UnreliableSend` and `LocalSend` could be conflated into one constructor. The only reason to have `UnreliableSend` is to spare the `NodeId` when sending the message remotely. `LocalSend` does not seem to use the `NodeId` for anything either.

### DIFF
--- a/src/Control/Distributed/Process.hs
+++ b/src/Control/Distributed/Process.hs
@@ -25,6 +25,7 @@ module Control.Distributed.Process
   , liftIO -- Reexported for convenience
     -- * Basic messaging
   , send
+  , usend
   , expect
   , expectTimeout
     -- * Channels
@@ -195,6 +196,7 @@ import Control.Distributed.Process.Serializable (Serializable)
 import Control.Distributed.Process.Internal.Primitives
   ( -- Basic messaging
     send
+  , usend
   , expect
     -- Channels
   , newChan
--- a/src/Control/Distributed/Process/Internal/Primitives.hs
+++ b/src/Control/Distributed/Process/Internal/Primitives.hs
@@ -255,11 +255,15 @@ send them msg = do
 unsafeSend :: Serializable a => ProcessId -> a -> Process ()
 unsafeSend = Unsafe.send
 
--- | Send a message without any reliability or ordering guarantees.
+-- | Send a message unreliably.
 --
 -- Unlike 'send', this function is insensitive to 'reconnect'. It will
 -- try to send the message regardless of the history of connection failures
 -- between the nodes.
+--
+-- Message passing with 'usend' is ordered for a given sender and receiver
+-- if the messages arrive at all.
+--
 usend :: Serializable a => ProcessId -> a -> Process ()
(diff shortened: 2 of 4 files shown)


hyperthunk:
This looks good to me. Just going to deal with some other PRs and get the CI build sorted and then I'll get this merged.

qnikst:
@hyperthunk can you force Travis to rerun this PR?

@facundominguez do you have some tips how to update semantics document to include `usend` there?

hyperthunk:
> @hyperthunk can you force Travis to rerun this PR?

Not unless the PR is updated. But I'll run the tests locally and merge anyway. I can't do any merges until the weekend now as I'm off-site and without my personal laptops until Friday night at the earliest.

facundominguez:
> @facundominguez do you have some tips how to update semantics document to include usend there?

The simplest is to introduce a set of unreliable messages in the *system* model. A rule for `usend` inserts them there, another rule drops a messages from the set (to model the fact they can be lost), and an extra rule for `expect` takes messages from this set in no particular order.

mboes:
You can rerun a build using the Travis CI web interface.

facundominguez:
Actually, I believe the current implementation is guaranteeing ordered delivery of unreliable messages. That is, if they arrive at all, they are received in the same order they were sent.

@hyperthunk, can you check this? If so, semantics could account for it as well.

hyperthunk:
Build failures were due to installation issues and it passed on the latest GHC. I'm inclined to merge.

hyperthunk:
@facundominguez I'll check. I cannot see why this wouldn't be the case though, except of course it is *only* delivery of unreliable messages that will be ordered. If you mingle `send` with `usend`, the ordering will be undefined.

facundominguez:
> ... it is only delivery of unreliable messages that will be ordered. (...)

Yeah, this is what I meant.

> If you mingle `send` with `usend`, the ordering will be undefined.

That's my understanding as well.

hyperthunk:
@facundominguez - well I think you're probably right, in so much as the ordering guarantee will hold between *two and only two* processes, since all `usend` traffic is routed via the node controller and the behaviour of the NC control channel should ensure that!

hyperthunk:
Do you fancy updating the semantics @facundominguez?

facundominguez:
Will propose an update to the semantics as soon as I have some time.

PR #174, opened 2015-01-17
Get the size of a process mailbox

qnikst:
I'm reopening DP-02 pull request now it contains a fully STM deque implementation.
I run benchmarks and didn't see any regressions, however better tests may be required.

@facundominguez, @edsko, please check this implementation if you have time.

### DIFF
--- a/src/Control/Distributed/Process/Internal/CQueue.hs
+++ b/src/Control/Distributed/Process/Internal/CQueue.hs
@@ -10,6 +10,7 @@ module Control.Distributed.Process.Internal.CQueue
   , enqueueSTM
   , dequeue
   , mkWeakCQueue
+  , queueSize 
   ) where
 
 import Prelude hiding (length, reverse)
@@ -17,10 +18,14 @@ import Control.Concurrent.STM
   ( atomically
   , STM
   , TChan
+  , TVar
+  , modifyTVar'
   , tryReadTChan
   , newTChan
+  , newTVarIO
   , writeTChan
   , readTChan
+  , readTVarIO
   , orElse
   , retry
   )
@@ -38,6 +43,7 @@ import Control.Distributed.Process.Internal.StrictList
   , append
   )
 import Data.Maybe (fromJust)
+import Data.Foldable (traverse_)
 import GHC.MVar (MVar(MVar))
 import GHC.IO (IO(IO))
 import GHC.Prim (mkWeak#)
@@ -46,19 +52,22 @@ import GHC.Weak (Weak(Weak))
 -- We use a TCHan rather than a Chan so that we have a non-blocking read
 data CQueue a = CQueue (StrictMVar (StrictList a)) -- Arrived
                        (TChan a)                   -- Incoming
+                       (TVar Int)                 -- Queue size
 
 newCQueue :: IO (CQueue a)
-newCQueue = CQueue <$> newMVar Nil <*> atomically newTChan
+newCQueue = CQueue <$> newMVar Nil <*> atomically newTChan <*> newTVarIO 0
 
 -- | Enqueue an element
 --
 -- Enqueue is strict.
 enqueue :: CQueue a -> a -> IO ()
-enqueue (CQueue _arrived incoming) !a = atomically $ writeTChan incoming a
+enqueue c !a = atomically (enqueueSTM c a)
 
 -- | Variant of enqueue for use in the STM monad.
 enqueueSTM :: CQueue a -> a -> STM ()
-enqueueSTM (CQueue _arrived incoming) !a = writeTChan incoming a
+enqueueSTM (CQueue _arrived incoming size) !a = do
+   writeTChan incoming a
+   modifyTVar' size succ
 
 data BlockSpec =
     NonBlocking
@@ -101,7 +110,7 @@ dequeue :: forall m a.
         -> BlockSpec         -- ^ Blocking behaviour
         -> [MatchOn m a]     -- ^ List of matches
         -> IO (Maybe a)      -- ^ 'Nothing' only on timeout
-dequeue (CQueue arrived incoming) blockSpec matchons = mask_ $
+dequeue (CQueue arrived incoming size) blockSpec matchons = mask_ $ checkDecrement $
   case blockSpec of
     Timeout n -> timeout n $ fmap fromJust run
     _other    ->
@@ -113,6 +122,11 @@ dequeue (CQueue arrived incoming) blockSpec matchons = mask_ $
                               -- no onException needed
          _other -> run
   where
+    checkDecrement f = do
+       mx <- f
+       mask_ $ traverse_ (const $ atomically $ modifyTVar' size pred) mx
+       return mx
+
     chunks = chunkMatches matchons
 
     run = do
@@ -245,5 +259,8 @@ dequeue (CQueue arrived incoming) blockSpec matchons = mask_ $
 
 -- | Weak reference to a CQueue
 mkWeakCQueue :: CQueue a -> IO () -> IO (Weak (CQueue a))
-mkWeakCQueue m@(CQueue (StrictMVar (MVar m#)) _) f = IO $ \s ->
+mkWeakCQueue m@(CQueue (StrictMVar (MVar m#)) _ _) f = IO $ \s ->
   case mkWeak# m# m f s of (# s1, w #) -> (# s1, Weak w #)
+
+queueSize :: CQueue a -> IO Int
+queueSize (CQueue _ _ size) = readTVarIO size
--- a/src/Control/Distributed/Process/Internal/CQueue.hs
+++ b/src/Control/Distributed/Process/Internal/CQueue.hs
@@ -16,46 +16,42 @@ module Control.Distributed.Process.Internal.CQueue
 import Prelude hiding (length, reverse)
 import Control.Concurrent.STM
   ( atomically
+  , check
   , STM
   , TChan
   , TVar
   , modifyTVar'
+  , writeTVar
   , tryReadTChan
   , newTChan
   , newTVarIO
   , writeTChan
   , readTChan
+  , readTVar
   , readTVarIO
   , orElse
   , retry
+  , registerDelay
   )
 import Control.Applicative ((<$>), (<*>))
-import Control.Exception (mask_, onException)
-import System.Timeout (timeout)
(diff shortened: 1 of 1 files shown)


hyperthunk:
Which benchmarks did we run? @mboes mentioned that Tweag may have some benchmarking code of their own which it would be good to run, even if it's not open-sourcable.

Would it not make sense to use `TMVar` here instead of a plain `TVar` for the arrived field? I suspect that might reduce the diff somewhat. Also I think we want to adapt `StrictMVar` to work with `TMVar` here, since the strictness semantics are the same for `MVar` and `TMVar` and there were good reasons why it was implemented for `MVar` in the first place (though I've lost the conversation I had with @edsko about that - we should probably chat with the well-typed folks about this on friend at some point).

Otherwise this looks reasonable to me though..

hyperthunk:
Also @qnikst - there are merge conflicts. Would you mind rebasing your local copy to resolve these and updating the commit range for the PR please?

qnikst:
We have `benchmarks` in distributed-process, I think that Latency benchmarks should be most affected. But more benchmarks will be useful to run.

About `TVar` vs `TMVar` in a situation when `dequeue` access is single threaded `TVar` is cheaper, in a concurrent calls both threads will continue computation, but only one will succeed, all other will be restared. `TMVar` will help with a case of concurrent calls to `dequeue` but I would argue that it still be not ideal as fairness of reads is not guaranteed. (But as we discussed in previous PRs we have single threaded access to dequeue). Also it simplifies TVar a bit so we don't need strict version, see below.

About strict version, I agree that having strict TVar will help from occasional problems, but here its safe as all calls are either `readTVar` or strict `modifyTVar'` and only `writeTVar` or `modifyTVar (const foo)` are problematic (if I missed such calls in my code, that it means its easy to make a mistake and I'll introcude strict TVar wrapper).

I'll be able to address all comments and rebase this or next night.

hyperthunk:
Ok sounds ok to me @qnikst. @facundominguez are you up for doing a quick peer review once the patch has been rebased please?

qnikst:
I have rebased PR. However I have not introduced strict version of `TVar` yet. `writeTVar` is used but only for result of `checkArrived`, if it's additional laziness can appear in that operation then I can introduce strict `TVar`.

src/Control/Distributed/Process/Internal/CQueue.hs
```diff
@@ -100,34 +105,57 @@ dequeue :: forall m a.
            CQueue m          -- ^ Queue
         -> BlockSpec         -- ^ Blocking behaviour
         -> [MatchOn m a]     -- ^ List of matches
-        -> IO (Maybe a)      -- ^ 'Nothing' only on timeout
-dequeue (CQueue arrived incoming) blockSpec matchons = mask_ $
+        -> IO (Maybe a)      -- ^ Nothing' only on timeout
```
facundominguez: Accidental apostrophe removal in comment.

src/Control/Distributed/Process/Internal/CQueue.hs
```diff
@@ -10,55 +10,60 @@ module Control.Distributed.Process.Internal.CQueue
...
 -- We use a TCHan rather than a Chan so that we have a non-blocking read
-data CQueue a = CQueue (StrictMVar (StrictList a)) -- Arrived
-                       (TChan a)                   -- Incoming
+data CQueue a = CQueue (TVar (StrictList a)) -- Arrived
+                       (TChan a)             -- Incoming
+                       (TVar Int)            -- Queue size
 
 newCQueue :: IO (CQueue a)
-newCQueue = CQueue <$> newMVar Nil <*> atomically newTChan
+newCQueue = CQueue <$> newTVarIO Nil <*> atomically newTChan <*> newTVarIO 0
```
facundominguez: `newTChanIO`? Just to avoid creating an unneeded transaction.

src/Control/Distributed/Process/Internal/CQueue.hs
```diff
@@ -100,34 +105,57 @@ dequeue :: forall m a.
...
+        -- stream. Having this transaction guarantees that all the
+        -- rest computation will not be rolled back if a new value
+        -- will enter the chan.
+        let loop = do
+              r <- tryReadTChan incoming
+              case r of
+                Nothing -> return ()
+                Just x  -> do modifyTVar' arrived (flip Snoc x)
+                              loop
+        atomically $ loop
```
facundominguez: There's a potential starvation of the reader here if writers are frequently modifying the `TChan`.

qnikst:
I definitely going to address comments about  missing `'` and `newTChanIO`, but the rest are quite more tricky. The `loop` is really a point for starvation, and there is more or less trivial fix, this transaction can be broken into multiple ones, so starvation will not be a problem anymore.
However similar problem may appear in another transaction, and the fix there is not so trivial,
so the question here do we need to care about:

a. multiple concurrent readers
b. concurrent reader and writer

if not I'd just add a documentation about the fact that implementation is suboptimal for those.

hyperthunk:
@facundominguez - thanks for the peer review!
@qnikst - with the loop, the *only* thing we need to consider is (b), concurrent reader and writer. There is only ever 1 reader (the process that owns this mailbox) and 1 writer (the node controller). I believe we're still vulnerable to reader starvation here, even with just 1 writer - do you both agree? What's the fix too?

qnikst:
>  I believe we're still vulnerable to reader starvation here, even with just 1 writer - do you both agree?

First loop fixed.

It's possible to implement same fix for another transaction.

However as it was discussed with @edsko  on #haskell-distributed channel, I'd like to have some more information on benchmarks before this will be merged. And in meantime will prepare a non stm solution (like it was before, @edsko is ok with that).

hyperthunk:
Ok @qnikst that sounds good. And yes, I agree more benchmarks would be a sensible prerequisite. What are we going to do about those?

qnikst:
I think wait for some time in order to find a good ideas how to construct benchmark could potentially show the difference. I can have access to Tweag I/O benchmarks but it's also not enough. I asked few other companies that build a high-performance solutions based on d-p, but they also do not have benchmarks yet.

qnikst:
And in the nearest future use https://github.com/haskell-distributed/distributed-process/pull/176 approach, as it doesn't change algorithm properties and is safe.

facundominguez:
@qnikst: Does this version of the PR address the negative queue sizes we found a while ago?

qnikst:
I think that negative queue size will not happen here, but fix for current version was merged. This PR was investigation of ways to implement dequeue completely in STM. But we have no benchmarks to show if this change is beneficial or not. I'm OK with closing PR.

src/Control/Distributed/Process/Internal/CQueue.hs
```diff
@@ -101,33 +106,58 @@ dequeue :: forall m a.
...
+        -- to break a transaction into multiple ones in order to
+        -- reduce number of retries. Transaction are split into
+        -- parts that are safe to use in concurrent enviroment in
+        -- presence of asynchonous exceptions.
+         
+        -- 1st transaction read consume all values from the input 
+        -- stream. Having this transaction guarantees that all the
+        -- rest computation will not be rolled back if a new value
+        -- will enter the chan.
+	let loop :: IO (IO ())
```
mboes: Why the extra monadic layer?

src/Control/Distributed/Process/Internal/CQueue.hs
```diff
@@ -101,33 +106,58 @@ dequeue :: forall m a.
...
-           goCheck chunks arr'
+    decrementJust f = do
+      mx <- f
+      when (isJust mx) (modifyTVar' size pred)
+      return mx
 
     waitChans ports on_block =
         foldr orElse on_block (map (fmap Just) ports)
 
+    run :: TVar Bool -> IO (Maybe a)
```
mboes: What does this change have to do with queue sizes?

qnikst: It should decrement queue size if value is read, and do nothing orderwise (I'm not sure if this PR worth reviewing, as anyway we have no benchmarks to understand if it's better than IO based or not, I think I should just close it until discussion will happen again)

src/Control/Distributed/Process/Internal/CQueue.hs
qnikst: I want to compose transaction, i.e. return next transaction outside of current one
(do I need to explain why do I need it, and why it's not possible to do inside one transaction?)
It's possible by wrapping value with a `Foldable` (or at least `Functor`?). Another way
to remove `IO` layer is to use `Maybe` instead, for example:
```
loop :: IO ()
loop = do
   let inner = atomically $ tryReadTChan incoming >>=
        traverse (\x -> modifyTVar' arrived (flip Snoc x) >> return inner)
   forM_ id inner
```

mboes: > I want to compose transaction, i.e. return next transaction outside of current one
Why?

Further, if that's what you're doing, then the comments are misleading, because the code below isn't the second transaction. It would the Nth+1 transaction where N is the number of items in the TChan.

Finally, sounds like you're overcomplicating it - if you want multiple transactions, then push the `atomically` further down, say around the `tryReadTChan`+`modifyTVar'`, which is in fact exactly what the existing code is doing?

qnikst: Yes, I need to update my comment above, it tells the reason why there are `N` transaction, but not 1, but it really misleading, as it states that all values are read.

I want to avoid long transactions, the longer transaction will be the more I need to pay for commit log, and the chance of retry will be larger. Suppose we have 2 threads one writing to channel and another reads from the channel (exactly the case here, and this is a reader). If any message will we written to the queue then my transaction will be retried, and this is definitely not what I want.

(Just to note once again, I'm really interested in discussing and clearing this PR, and using best practices here, but the chance that it will be merged without much work around benchmarks is low, and current benchmarks shows that solutions are on par)

hyperthunk:
Are we in a position to benchmark this now?

qnikst:
I'm not sure about best way here, ideally we need to run some programs that heavily uses different sort of receive and it's a bottleneck. At this point I don't have such programs, and I'm not sure if I know how to write good benchmark.

hyperthunk:
I have the same thoughts. I think we should leave this for now. Maybe add another label, `parked` or some such?

facundominguez:
Feel free to resubmit.

PR #176, opened 2015-01-19
DP-02: simple approach for maintaining queue size.

qnikst:
This commit introduces basic approach to maintain size of the
queue by keeping TVar Int in the state, then each writer will
increment this state, and deque will decrement it if and only
if it outputs value.

### DIFF
--- a/src/Control/Distributed/Process/Internal/CQueue.hs
+++ b/src/Control/Distributed/Process/Internal/CQueue.hs
@@ -10,6 +10,7 @@ module Control.Distributed.Process.Internal.CQueue
   , enqueueSTM
   , dequeue
   , mkWeakCQueue
+  , queueSize 
   ) where
 
 import Prelude hiding (length, reverse)
@@ -17,10 +18,14 @@ import Control.Concurrent.STM
   ( atomically
   , STM
   , TChan
+  , TVar
+  , modifyTVar'
   , tryReadTChan
   , newTChan
+  , newTVarIO
   , writeTChan
   , readTChan
+  , readTVarIO
   , orElse
   , retry
   )
@@ -38,6 +43,7 @@ import Control.Distributed.Process.Internal.StrictList
   , append
   )
 import Data.Maybe (fromJust)
+import Data.Foldable (traverse_)
 import GHC.MVar (MVar(MVar))
 import GHC.IO (IO(IO))
 import GHC.Prim (mkWeak#)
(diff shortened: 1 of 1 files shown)


qnikst:
This is non-STM version, that is also correct. I think that it should be a way forward at least in the short-term.

src/Control/Distributed/Process/Internal/CQueue.hs
```diff
@@ -113,6 +122,11 @@ dequeue (CQueue arrived incoming) blockSpec matchons = mask_ $
                               -- no onException needed
          _other -> run
   where
+    checkDecrement f = do
+       mx <- f
+       mask_ $ traverse_ (const $ atomically $ modifyTVar' size pred) mx
```
facundominguez: This `mask_` is nested with the `mask_` of `dequeue`. Maybe it is not needed.

qnikst:
What will be our actions here?

hyperthunk:
> What will be our actions here?

I'm not sure there's anything left to do. I still feel like there are some risks (due to lack of atomicity when combing `MVar` and STM based code like this, but I can't call them out explicitly right now. I will extend the discussion before merging (over the weekend) if I feel it's necessary.

src/Control/Distributed/Process/Internal/CQueue.hs
```diff
@@ -101,7 +109,7 @@ dequeue :: forall m a.
         -> BlockSpec         -- ^ Blocking behaviour
         -> [MatchOn m a]     -- ^ List of matches
         -> IO (Maybe a)      -- ^ 'Nothing' only on timeout
-dequeue (CQueue arrived incoming) blockSpec matchons = mask_ $
+dequeue (CQueue arrived incoming size) blockSpec matchons = mask_ $ decrementJust $
```
hyperthunk: Surely we should *only* decrement if the dequeue operation does not time out!?

qnikst: on a timeout `Nothing` is returned, so we don't decrement then. And because of `mask_` timout may happen only when thread is blocked on `MVar` so we are happy in any case, I suppose.

hyperthunk: Oh gosh yes, sorry I didn't read the code properly to realise that `decrementJust` takes the result of evaluating the code it's given as its input. That's fine then. And I agree (I think) about the use of `mask_` here, Edsko's previous comments about uninterruptible masking notwithstanding...

qnikst: I had a conversation with Edsko about mask usage here, and he said he is fine with that. I can try to find logs if needed.

hyperthunk:
I'm keen to merge this. Are we happy that there aren't any significant performance regressions?

qnikst:
I don't have benchmarks rather that in `benchmarks` folder, but those shows no performance regressions.

hyperthunk:
Ok good. I think we'll risk it. We can always bisect if regressions are reported in future.

hyperthunk:
We should go and update `getProcessInfo` to include the queue length now! :)

qnikst:
Will do today.

hyperthunk:
Awesome!

PR #181, opened 2015-02-13
Implement `Process` `MonadFix` instance

qnikst:


qnikst:
`MonadFix` is useful when distributed-process is used together with A-FRP libraries, for example netwire.

PR #182, opened 2015-02-19
Return size of the queue in ProcessInfo.

qnikst:


hyperthunk:
Looks good. Why don't we change the type of queueSize so it's no longer in `Maybe` as we're always able to calculate that now?

qnikst:
Reasonable, I have tried to not break API at all. But I can change to of cause.

hyperthunk:
Yes, let's break it.

qnikst:
Updated

PR #184, opened 2015-02-25
Implement monitoring for registry values

qnikst:
Implement remote Process monitoring as Agent.

### DIFF
--- a/src/Control/Distributed/Process/Node/RegistryAgent.hs
+++ b/src/Control/Distributed/Process/Node/RegistryAgent.hs
@@ -1,0 +1,80 @@
+{-# LANGUAGE RecursiveDo  #-}
+-----------------------------------------------------------------------------
+---- |
+---- Module      :  Control.Distributed.Process.Node.RegistryAgent
+---- Copyright   :  (c) Tweag I/O 2015
+---- License     :  BSD3 (see the file LICENSE)
+----
+---- Maintainer  :  Tim Watson <watson.timothy@gmail.com>
+---- Stability   :  experimental
+---- Portability :  non-portable (requires concurrency)
+----
+---- This module provides a registry monitoring agent, implemented as a
+---- /distributed-process Management Agent/. Every 'node' starts this agent on
+---- startup. The agent will monitor every remote process that was added to the
+---- local registry, so the node removes the process from the registry when it
+---- dies or when a network failure is detected.
+----
+-------------------------------------------------------------------------------
+
+module Control.Distributed.Process.Node.RegistryAgent
+    ( registryMonitorAgent
+    ) where
+
+import Control.Distributed.Process.Management
+import Control.Distributed.Process.Internal.Types
+import Control.Distributed.Process.Internal.Primitives
+import Data.Foldable (forM_)
+import Data.Map (Map)
+import qualified Data.Map as Map
+
+registryMonitorAgentId :: MxAgentId
+registryMonitorAgentId = MxAgentId "service.registry.monitoring"
+
+-- | Registry monitor agent
+--
+-- This agent listens for 'MxRegistered' and 'MxUnRegistered' events and tracks
+-- all remote 'ProcessId's that are stored in the registry.
+--
+-- When a remote process is registered, the agent starts monitoring it until it
+-- is unregistered or the monitor notification arrives.
+--
+-- The agent keeps the amount of labels associated with each registered remote
+-- process. This is necessary so the process is unmonitored only when it has
+-- been unregistered from all of the labels.
+--
+registryMonitorAgent :: Process ProcessId
+registryMonitorAgent = do
+  nid <- getSelfNode
+  -- For each process the map associates the 'MonitorRef' used to monitor it and
+  -- the amount of labels associated with it.
+  mxAgent registryMonitorAgentId (Map.empty :: Map ProcessId (MonitorRef, Int))
+    [ mxSink $ \(ProcessMonitorNotification _ pid _) -> do
+        mxUpdateLocal (Map.delete pid)
+        mxReady
+    , mxSink $ \ev -> do
+        case ev of
+          MxRegistered pid _
+            | processNodeId pid /= nid -> do
+            hm <- mxGetLocal
+            m <- liftMX $ mdo
+              let (v,m) = Map.insertLookupWithKey (\_ (m',r) _ -> (m',r+1))
+                                                  pid (mref,1) hm
+              mref <- maybe (monitor pid) (return . fst) v
+              return m
+            mxSetLocal m
+          MxUnRegistered pid _
+            | processNodeId pid /= nid -> do
+            hm <- mxGetLocal
+            forM_ (pid `Map.lookup` hm) $ \(mref, i) ->
+              let i' = pred i
+              in if i' == 0
+                 then do liftMX $ unmonitorAsync mref
+                         mxSetLocal $! pid `Map.delete` hm
+                 else mxSetLocal $ Map.insert pid (mref,i') hm
+          _ -> return ()
+        mxReady
+      -- remove async answers from mailbox
+    , mxSink $ \RegisterReply{} -> mxReady
+    , mxSink $ \DidUnmonitor{} -> mxReady
+    ]
--- a/distributed-process.cabal
+++ b/distributed-process.cabal
@@ -76,6 +76,7 @@ Library
                      Control.Distributed.Process.Internal.WeakTQueue,
                      Control.Distributed.Process.Management,
                      Control.Distributed.Process.Node,
+                     Control.Distributed.Process.Node.RegistryAgent
                      Control.Distributed.Process.Serializable,
                      Control.Distributed.Process.UnsafePrimitives
                      Control.Distributed.Process.Management.Internal.Agent,
--- a/src/Control/Distributed/Process/Internal/Primitives.hs
+++ b/src/Control/Distributed/Process/Internal/Primitives.hs
@@ -72,6 +72,7 @@ module Control.Distributed.Process.Internal.Primitives
   , unlink
   , monitor
   , unmonitor
+  , unmonitorAsync
   , withMonitor
     -- * Logging
   , say
--- a/src/Control/Distributed/Process/Management/Internal/Types.hs
+++ b/src/Control/Distributed/Process/Management/Internal/Types.hs
@@ -30,6 +30,7 @@ import Control.Distributed.Process.Internal.Types
   , NodeId
   )
 import Control.Monad.IO.Class (MonadIO)
+import Control.Monad.Fix (MonadFix)
 import qualified Control.Monad.State as ST
   ( MonadState
   , StateT
@@ -123,6 +124,7 @@ newtype MxAgent s a =
   } deriving ( Functor
              , Monad
              , MonadIO
+             , MonadFix
              , ST.MonadState (MxAgentState s)
              , Typeable
              , Applicative
--- a/src/Control/Distributed/Process/Node.hs
+++ b/src/Control/Distributed/Process/Node.hs
@@ -68,6 +68,7 @@ import Control.Exception
   )
 import qualified Control.Exception as Exception (Handler(..), catches, finally)
 import Control.Concurrent (forkIO, forkIOWithUnmask, killThread, myThreadId)
+import Control.Distributed.Process.Node.RegistryAgent (registryMonitorAgent)
 import Control.Distributed.Process.Internal.StrictMVar
   ( newMVar
   , withMVar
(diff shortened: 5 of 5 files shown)


qnikst:
Associated pull request to distributed-process-tests: https://github.com/haskell-distributed/distributed-process-tests/pull/15

qnikst:
@facundominguez please recheck my code when you'll have time.

src/Control/Distributed/Process/Node/RegistryAgent.hs
```diff
@@ -0,0 +1,81 @@
...
+                    MxUnRegistered pid _
+                        | processNodeId pid /= nid -> do
+                        hm <- mxGetLocal
+                        forM_ (pid `Map.lookup` hm) $ \(mref, i) ->
+                           let !i' = pred i
+                           in if i' == 0
+                              then do liftMX $ unmonitorAsync mref
+                                      mxSetLocal $! pid `Map.delete` hm
+                              else mxSetLocal $ Map.insert pid (mref,i') hm
+                    _ -> return ()
```
facundominguez: The alternatives of this case can be rewritten to use `Map.insertLookupWithKey` and  `Map.updateLookupWithKey` to do a single traversal. The effectful computation can be decided by examining the result.

src/Control/Distributed/Process/Node/RegistryAgent.hs
```diff
@@ -0,0 +1,81 @@
...
+import Control.Distributed.Process.Internal.Types
+import Control.Distributed.Process.Internal.Primitives
+import Data.Foldable (forM_)
+#if MIN_VERSION_containers(5,0,0)
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+#else
+import Data.Map (Map)
+import qualified Data.Map as Map
+#endif
```
facundominguez: Scrictness is given up in the older versions. Isn't it better to use the lazy version always and strictify the code below if needed at some place? I cannot spot any place where it is needed though.

qnikst: Really it's not needed that both keys and values arrive in even `NF`.

src/Control/Distributed/Process/Node/RegistryAgent.hs
qnikst: Unfortunately both of those will not work here.

1. `insertLookupWithKey` will not work as I have to have `a` that is `(MonitorRef, Int)` and I need to create `MonitorRef` only if there is no value. 

2. `updateLookupWithKey` returns either old value if value was deleted or changed value, so I have no chance
   to understand what happens as value `(mon,1)` can refer either to changed value or to removed.

hyperthunk:
This is looking good, just working through the tests.

facundominguez:
LGTM

qnikst:
Yes, also I'm planning to merge this after #199. To make integration process more clean.

qnikst:
Prepared for a merge.

src/Control/Distributed/Process/Node/RegistryAgent.hs
```diff
@@ -0,0 +1,81 @@
...
+
+-- | Registry monitor agent
+--
+-- This agent listens for 'MxRegistered' and 'MxUnRegistered' events and tracks
+-- all remote 'ProcessId's that are stored in the registry.
+--
+-- When a remote process is registered, the agent starts monitoring it until it
+-- is unregistered or the monitor notification arrives.
+--
+-- The agen keeps the amount of labels associated with each registered remote
```
mboes: typo

mboes:
Can we get a link to the actual issue? I have no idea what DP-100 means. Nor the context for why this needs to go in d-p as opposed to where the source code originally came from (where?). Update the title of this PR accordingly. "Dp 100" means nothing.

facundominguez:
https://cloud-haskell.atlassian.net/browse/DP-100

qnikst:
This is issue #212

src/Control/Distributed/Process/Node/RegistryAgent.hs
```diff
@@ -0,0 +1,81 @@
...
+-- The agent keeps the amount of labels associated with each registered remote
+-- process. This is necessary so the process is unmonitored only when it has
+-- been unregistered from all of the labels.
+--
+registryMonitorAgent :: Process ProcessId
+registryMonitorAgent = do
+  nid <- getSelfNode
+  -- For each process the map associates the 'MonitorRef' used to monitor it and
+  -- the amount of labels associated with it.
+  mxAgent registryMonitorAgentId (Map.empty :: Map ProcessId (MonitorRef, Int))
```
mboes: Any downside to simply having as many `MonitorRef`'s as there are registered labels, rather than reference counting? I don't think the memory saving (if any) is worth it, seeing as multiple labels for the same `ProcessId` is not the common case. Dropping the reference counting would make the code simpler, I reckon.

src/Control/Distributed/Process/Node/RegistryAgent.hs
```diff
@@ -0,0 +1,81 @@
+{-# LANGUAGE BangPatterns #-}
```
mboes: A single module for a single function sounds overkill. I'd inline `registryMonitorAgent` into `Node.hs`.

src/Control/Distributed/Process/Node/RegistryAgent.hs
qnikst: Yes, that will work, the differences are memory footprint and complexity of the `MxUnRegistered` case, that will be slightly higher (we will need traverse the structure and remove `MonitorRef`). `MxRegistered` will be simplified, another one will be more or less the same.
So I'm open to this change.

@facundominguez, any thoughts here?

facundominguez: I'm not convinced the rewriting is worth the trouble. We already have a reasonable PR. Your call.

mboes: > complexity of `MxUnRegistered` case, that will be slightly higher (we will need traverse the structure and remove `MonitorRef`).

I don't see it. The following is untested but compiles:

```Haskell
        case ev of
          MxRegistered pid _
            | processNodeId pid /= nid -> do
              mref <- liftMX $ monitor pid
              mxUpdateLocal (Map.insert pid mref)
          MxUnRegistered pid _
            | processNodeId pid /= nid -> do
              mrefs <- mxGetLocal
              forM_ (pid `Map.lookup` mrefs) $ \mref -> do
                liftMX $ unmonitorAsync mref
                mxUpdateLocal (Map.delete pid)
          _ -> return ()
```
I expect the memory usage to be pretty much the exactly the same as with reference counting: monitors are cheap, and registering a process under multiple labels is uncommon anyways.

qnikst: The code above do not solve the problem that original version intended to solve, i.e. if process is registered multiple times but unregistered only once then it will be removed from map, but it shouldn't. Also I'm not terribly happy about "leaking" mrefs. I have to check code in order to understand if its safe to do. Also here we have a monitor call for each registered process (I can agree that it's a rare case).

mboes: Oops - late at night pasted wrong code snippet. I meant to paste this one:

```Haskell
        case ev of
          MxRegistered pid label
            | processNodeId pid /= nid -> do
              mref <- liftMX $ monitor pid
              mxUpdateLocal (Map.insert label mref)
          MxUnRegistered pid label
            | processNodeId pid /= nid -> do
              mrefs <- mxGetLocal
              forM_ (label `Map.lookup` mrefs) $ \mref -> do
                liftMX $ unmonitorAsync mref
                mxUpdateLocal (Map.delete label)
          _ -> return ()
```
i.e. as the previous comment was saying, one monitor ref per label, not per process. The only difference with the previous snippet is the key of the map. Everything else is the same. BUT, there is a complication, as you point out. It's just that it's not in the `MxUnregistered` case but in the `ProcessMonitorNotification` case, which when written out becomes something like:

```Haskell
mxSink $ \(ProcessMonitorNotification mref _ _) -> do
        mxUpdateLocal $ Map.filter (== mref)
        mxReady
```

I assume you meant to write `ProcessMonitorNotification` instead of `MxUnregistered`. The code is just as simple, but note the linear operation, which I also assume is what you were referring to. It's the same linear operation that's already [here](https://github.com/haskell-distributed/distributed-process/blob/master/src/Control/Distributed/Process/Node.hs#L812), over the exact same set of labels. Processes shouldn't be overusing the registry to register thousands of processes, so that's just about ok.

qnikst: I was thinking about `Monoid m => Map ProcesaId (m MonitorRef)` in that case complexity is  in `MxUnregister` because delete is linear or logarithmic there. So I think I was correct but thought about different solution in design space.
About this code we need to check that reregistration `MxUnregistred` then `MxRegistered` exactly in this order, otherwise `MxRegistered` case is not correct, but I hope that ordering and existence of messages are provided.

mboes: Well, if events get dropped then even the current code will have trouble. Events on the management bus shouldn't get dropped.

If we must assume that messages do get reordered (because the Management API documentation claims to provide no guarantees), or dropped or duplicated, then I'm not sure that using the Management API for building this functionality is a great option after all.

As @facundominguez says, not worth agonizing over the implementation of this function, since it's basically correct as-is. Though on the assumption that the Management API does provide _some_ guarantees. @hyperthunk could you confirm what those are precisely?

qnikst:
I've addressed comment about BangPatterns, about guarantees that we have, it seems that a. messages couldn't be lost, b. MxUnregistered message will be missing on reregistering:

https://github.com/haskell-distributed/distributed-process/blob/master/src/Control/Distributed/Process/Node.hs#L857-L874

So I want to introduce test for that issue and check if it's a case, or I just miss some bits in code.

mboes:
@qnikst I don't think this is a test issue. It's a design issue. We can't assume make assumptions based on the current implementation of an API - we can only make the assumptions that the documentation for that API allows us to make. This PR is currently blocked on feedback from @hyperthunk about the intended guarantees of the Management API. Actually this question should probably be its own Github issue, with this PR marked as blocked on that.

mboes:
@qnikst though, rereading your comment - it may be that you've uncovered a bug, independent of the design level discussion about the guarantees of the Management API.

qnikst:
Blocked on #240 , #241

hyperthunk:
@mboes is it just the blocking issues that you believe are a bug in the Management API? I'm going to try and address these with @qnikst.

hyperthunk:
Guys... Can this be merged once I've resolved #241, which I plan to address today?

qnikst:
Yes, I can rebase everything on the top of the master and check that CI passes, then this PR may be merged. Any objections?

mboes:
Could this either, be folded into `C.D.P.Node` (ideally - do we really need a whole new module just for one definition?) or moved to the `Internal.*` namespace? AFAIU this agent is just yet another "service process" internal to the Node Controller.

Other than that, no objection.

qnikst:
Really we do not have a common policy about where agents should be put. I'm ok with moving that to C.D.P.Node.

mboes:
In this case - isn't this a "service process" like any other? If so it should just go where the other processes are.

qnikst:
Technically it's not a ordinary process, but I agree with this reasoning.

hyperthunk:
I'm okay with it living in Node.hs too, or being internal. The code in Node.hs could get a bit cluttered if we keep adding to it. Note that the implementation of the other service processes is currently scattered - the simple built in logger (that backs `say`) resides in Node.hs, but all the management, tracing, and debugging code resides elsewhere.

hyperthunk:
Ok I have something to add to this... I am working on improving and fixing the management event bus. Currently some events aren't firing, or could potentially fire in the wrong order (I.e., receive before send - although we don't guarantee causal ordering, for MxSend and MxReceived in the case of two local processes, we *can* guarantee it, and there I believe we should endeavour to do so).

Anyway, in the process of pulling that path together, I realise that now the management event bus is designed the way it is, the complex trace controller process can be rewritten as a management agent, and we will remove a lot of garbage from the code base by doing that, and simplify the code a lot.

So I wonder if we should have `Control.Distributed.Process.Internal.SystemProcesses` (or multiple modules beneath that?) and put the `say` logger and your registry monitoring agent in there? And export `startSystemProcesses` from there, so the node controller remains ignorant and unaffected in future?

hyperthunk:
CI failed, but only for GHC 7.1 - do we really want to support that far back? Also, I suspect we need to rebase with master here.

mboes:
There's no such thing as 7.1. I think this a case of writing `- 7.10` instead of `- '7.10'` in travis.yml...

hyperthunk:
Ok @mboes that makes sense now. @qnikst I also need to fix reregister semantics to fire unregistered followed by registered events

qnikst:
Closed in favor of #322

PR #185, opened 2015-03-02
Workaround for throwTo problem that can potentially block.

qnikst:


hyperthunk:
Why do we need this to be asynchronous? I'm not opposed to it, just want to understand the rationale.

qnikst:
Otherwise node controller could get stuck if process in uninterruptible operation. See https://cloud-haskell.atlassian.net/browse/DP-105

mboes:
I'm not sure what this patch is trying to fix - DP105 shows a contrived example of how *not* to use `uninterruptibleMask`, which is one of the few functions in base that comes with a big warning in all caps:

"Like mask, but the masked computation is not interruptible (see Control.Exception). THIS SHOULD BE USED WITH GREAT CARE, because if a thread executing in uninterruptibleMask blocks for any reason, then the thread (and possibly the program, if this is the main thread) will be unresponsive and unkillable. This function should only be necessary if you need to mask exceptions around an interruptible operation, and you can guarantee that the interruptible operation will only block for a short period of time."

http://hackage.haskell.org/package/base-4.7.0.2/docs/Control-Exception.html

I don't recommend making abuse of `uninterruptibleMask` any easier. In fact one could argue that according to the principle of least suprise, `exit selfPid ()` ought to behave exactly like `throwTo selfTid ()`, which with this patch it no longer does (e.g. if the process is currently masking exceptions).

facundominguez:
> which with this patch it no longer does

Without this path, exit is asynchronous as well.

You don't need to use `uninterruptibleMask`, have the worker thread use a foreign call which never returns instead.

qnikst:
Without this patch `exit` was asynchronous on the contrary to throwTo, because exit sends a message to not and immediately exits, and this is a node who locks, so no semantic change here.

Unfortunately there are valid uses of uninterruptibleMask and it would be OK if user process locks but not node.

qnikst:
not-returning `uninterruptibleMask` or foreign call is definitely an extremal case, and if such situation happens this is effectively design flaw in user application. However there are use cases where a problem may occur, for example `uninterruptibleMask` or non-allocating loop (this usually happens in FFI polling functions, and I know that Tweag had such examples in related libraries). In all those cases exception will be eventually delivered but the node will be blocked until it happens. This will  lead to a poor behaviour like hardly tracked lags.

One can say that it's a user responsibility to write code in safe way so node will not be confused, one the other current behaviour make C-H node vulnerable to user error, and this makes it hardy coupled with user code. This should have value if C-H user wants to run 3-rd party Processes in his node.

mboes:
Ah I hadn't realized `exit` defers to the node controller no matter what. Nevertheless the point remains...

> You don't need to use uninterruptibleMask, have the worker thread use a foreign call which never returns instead.

Also something that is called out explicitly in the standard library documentation, and the reason why GHC supports interruptible foreign calls.

http://hackage.haskell.org/package/base-4.7.0.2/docs/Control-Exception.html#v:throwTo

... so then, what *legitimate*, *concrete* use case is this patch designed to fix? I don't feel strongly either way - I just tend to prefer not adding any additional indirections (and potentially some overhead) just to work around code that is arguably broken. Especially indirections that affect the relative ordering of exception delivery wrt other events (I don't know whether this is allowed or disallowed by the semantics). Now if we can find good concrete examples, that's a different matter. :)

qnikst:
Does my example with non-allocating loop counts? 

I.e. 
```
spawnLocal node  (liftIO $ loopWhileM (readIORef shouldStop) $ pollWithEventData someaction) 
```

One point that this code may introduce a bad behaviour even without cloud haskell, but existanse of additional block makes things harder. And there is a workaround - have a process that is a wrapper. So for example: `callLocal (liftIO $ ...)` will be not affected by this problem.

> Especially indirections that affect the relative ordering of exception delivery wrt other events (I don't know whether this is allowed or disallowed by the semantics).

I'd say it doesn't introduce any reordering as currently message delivery (moment when process will read it from his mailbox) and exception delivery are not ordered in any way.

mboes:
Non allocating loops are a more interesting use case in my mind (previous
message sent before I saw this one).
​
> I'd say it doesn't introduce any reordering as currently message delivery
(moment when process will read it from his mailbox) and exception delivery
are not ordered in any way.

Indeed, but currently by the time the node controller proceeds to
dispatching the next event, it knows that the target thread has been
interrupted (mind, not yet killed). I don't know if a process is allowed to
exploit that information.

Ultimately, I for one am feeling increasingly neutral about this. As you
say @qnikst, this comes down to whether we believe we ought to exploit an
assumption of exception hygiene in user code, or not.

qnikst:
> Indeed, but currently by the time the node controller proceeds to dispatching the next event, it knows that the target thread has been interrupted (mind, not yet killed). I don't know if a process is allowed to exploit that information.

Interesting point, I'm not sure how this information can be used, but I need to think more about this.

From my side I can just write down pros and cons that I see, and pass a decision to others :)

pros:

1. code becomes less vulnerable for user hardly tracked user errors that includes situation when process doesn't enter safe points for a long time. 

cons: 

1. it introduces unnecessary overhead for common case. (There is a possible improvement - node can prefork a number of processes and pass an action to them, this will reduce overhead).

2. it opens a window for DDOS attack - if one process will spam another (that refuses to die) with exceptions, node may generate too much processes. (An improvement for cons 1, also heals this  one).

facundominguez:
ehem, related: There is also a potential deadlock in `forkProcess` which uses uninterruptibleMask while trying to acquire the same MVar that `exit` is acquiring before blocking in `throwTo` (`localState node`). 

To solve this the `throwTo` call should be moved at least out of the `modifyMVar` closure in `exit`.

qnikst:
@facundominguez do I understand correctly that it's not a case if we are forking, because then `withLocalProc` immediately exits?

facundominguez:
Right.

facundominguez:
Reported issue as https://cloud-haskell.atlassian.net/browse/DP-106.

hyperthunk:
I've responded to the ticket, thanks @facundominguez. So I'm *really* not sure what we're advocating here now. Are we _just_ making sure the NC doesn't block on throwTo? I think the fix I've suggested on Jira (which is to let all the state cleanup happen in the terminated process) would be cleaner and avoids the need to forkIO for each exception we want to throw.

Also all this stuff about the asynchronicity (or otherwise) of exit is really confusing. The exit primitive is meant to be async, but we're not talking about that here right? And from whose perspective async?

qnikst:
@hyperthunk this PR and discussion is about current behaviour of NC that could be blocked (become unresponsive) when it transfers and exception (via `exit` or `kill`). This behaviour may play in following situations:

1. user uses `uninterruptibleMask` (attached in DP-105);
2. capability where thread is located was blocked by unsafe FFI code;
3. exeption can't be received because process is in non-allocating loop;

All code situations above may be considered as a bad user code, and have workarounds in client library side. However the question in this PR is the following: *do we want NC to be resilient to such problems*? If we do then we should consider merging this PR (or maybe some modification if there are  problems with this code). If we don't want complicate NC because of this possible problem then we need to document it.

So let me repeat major question for this PR: do we want NC to be resilient to such problems?

mboes:
@hyperthunk for some context, the issue is that in GHC exceptions delivered
via `throwTo' are asynchronous exceptions, but they are *delivered*
synchronously. That is, the thread calling `throwTo` (i.e. the NC) blocks
until the target thread has had the async exception delivered to it. Hence
why users should not prevent delivery of async exeptions for any extended
period of time, e.g. using `uninterruptibleMask`. This PR posits that d-p
ouught to be better behaved in the face of user foolishness. I for one am
undecided about whether this is the right thing to do, but it's worth
noting that as @facundominguez points out, it's surprisingly easy to be
foolish: just write some code that runs in a tight loop for a long time
without allocating.

`exit` itself is completely async either way, yes.

facundominguez:
> I've responded to the ticket, thanks @facundominguez

I think you refer to DP-106. This PR is more related to https://cloud-haskell.atlassian.net/browse/DP-105. There is a test example and some more discussion there, including what I think a legit use of uninterruptibleMask_ in the second comment.

hyperthunk:
I'm still worried about the overheads. But in answer to the main question here, yes, I think we do want the NC to be resilient and not allow itself to get blocked. Let me start reviewing this in more detail now I'm back in the UK.

facundominguez:
Here there is another situation that can block the NC: https://cloud-haskell.atlassian.net/browse/DP-110

qnikst:
I've squashed changes. Any objections to include this?

mboes:
I'm not opposed to seeing this merged.

qnikst:
Thanks. Merging than. If anybody will be able to find better solution here we could adopt that.

PR #187, opened 2015-03-06
Stop all processes in closeLocalNode.

qnikst:


qnikst:
Still experience some deadlocks, I'll have to recheck.

hyperthunk:
Yes, just one exception please.

qnikst:
Ok, but there is a problem that process could recover from exception, however it should not recover from `ProcessKilled` one, and thus process could not terminate.

qnikst:
Updated, only one exception is sent to the process, also I left `join` trick in order to kill new processes that were spawned after call to `closeLocalNode`.

src/Control/Distributed/Process/Node.hs
```diff
@@ -319,11 +322,38 @@ startServiceProcesses node = do
...
+-- or being able to gracefully shutdown.
 closeLocalNode :: LocalNode -> IO ()
-closeLocalNode node =
-  -- TODO: close all our processes, surely!?
-  NT.closeEndPoint (localEndPoint node)
+closeLocalNode node = join $ terminateProcesses
+  where
+    terminateProcesses :: IO (IO ())
+    terminateProcesses = do
+      st <- modifyMVar (localState node) (\st -> return (st, st))
```
mboes: Or much more simply, just `readMVar`?

src/Control/Distributed/Process/Node.hs
```diff
@@ -319,11 +322,38 @@ startServiceProcesses node = do
...
+      let mxACPid = case localEventBus node of
+                      MxEventBus pid _ _ _ -> pid
+                      MxEventBusInitialising -> error "terminateLocalProcesses: The given node is not initialized."
```
mboes: Is reaching this branch a bug, i.e. an impossible branch? Or is this a possible race condition?

src/Control/Distributed/Process/Node.hs
```diff
@@ -319,11 +322,38 @@ startServiceProcesses node = do
...
+          pids = filter (/=mxACPid) $ map processId $ Map.elems (st ^. localProcesses)
+      if pids == []
+        then do runProcess node $ do
+                  -- Trying to kill the management agent controller prevents other processes
+                  -- from terminating.
+                  mapM_ monitor pids
+                  mapM_ (flip exit "closing node") pids
```
mboes: We should be using `kill`, not `exit`, as in fact is consistent with your proposed docstring for this function which says "kill".

src/Control/Distributed/Process/Node.hs
```diff
@@ -319,11 +322,38 @@ startServiceProcesses node = do
...
+                  replicateM_ (length pids) $ receiveWait
+                    [ match $ \(ProcessMonitorNotification _ _ _) -> return () ]
+                return (join terminateProcesses)
+        else return stopNC
+
+    stopNC =
```
mboes: `stopNC`  ought to be made a top-level definition for shutting down the node controller, to avoid code duplication and avoid getting it wrong in some places over time.

src/Control/Distributed/Process/Node.hs
```diff
@@ -319,11 +322,38 @@ startServiceProcesses node = do
...
+-- process is not guaranteed to continue. So this is application
+-- responsibility to either close all it's processes before dying
+-- or being able to gracefully shutdown.
 closeLocalNode :: LocalNode -> IO ()
-closeLocalNode node =
-  -- TODO: close all our processes, surely!?
-  NT.closeEndPoint (localEndPoint node)
+closeLocalNode node = join $ terminateProcesses
+  where
+    terminateProcesses :: IO (IO ())
```
mboes: Why the potentially endless loop hammering processes `exit` exceptions? Surely one only gets a `ProcessMonitorNotification` after the process in fact no longer exists? If not, that's a bug, which I don't think it's justified to paper over by looping in this function.

mboes:
This change would moreover warrant an accompanying test.

src/Control/Distributed/Process/Node.hs
qnikst: In a new version it doesn't, because this function doesn't start new iteration unless all processes from current generation were killed. this means that for each process there will be only one exception. and then when all processes were killed then there will be a new check if there are new processes. If so then termination procedure continues. There are few other possible solutions:
1. Invent new procedure that will not allow staring new processes when node is closing
2. Use agent as agent has information about processes born and death.

src/Control/Distributed/Process/Node.hs
qnikst: Taken from original code, I agree that it should be simplified.

src/Control/Distributed/Process/Node.hs
qnikst: This is possible race condition between node close and initialization.

src/Control/Distributed/Process/Node.hs
qnikst: Agree, taken from original code without additional check.

src/Control/Distributed/Process/Node.hs
qnikst: Agree.

qnikst:
All tests are ready but I need to review them myself PR will be opened this night.

qnikst:
I have addressed all comments except request for additional test (as it will be done later and linked to this PR) and  comment about potential loop.

src/Control/Distributed/Process/Node.hs
qnikst: About why this solution was chosen. Currently we have following solutions:

1. original - same as this one but without looping at all. Basically this means that if any process will be spawned after start of `terminationProcess` they will survive. This solution may block any of process will refuse to die and may exit even if there are existing processes with a very large time frame for premature exit

2. my first attempt - had a loop but had no monitoring notifications. This solution had a tiny frame for prepature exit, but could loop if this process is in race with self restarting process. I.e. process that traps kill and spawn new process.

3. my second attempt (current one) - this is the same as 1 but with looping. Such looping reduces time frame for premature exit (and basically could entirely remote it), but can potentially block if there is self-restarting process.

Summing all this up. I think that we need to either decide to implement some heavyweight solution, or 3 could be the best fit.

mboes: Good point about new processes, but the current solution is insufficient in addition to being dangerous: the fact that there are no processes any longer on the node does not mean that a process on another node can't spawn a process on this node before the node is actually shut down.

The proper fix in my mind is neither of the 3 solutions above: clearly nodes need one additional state, say a "closing" state. When a node is in the closing state, no process is allowed to assign any new resources (e.g. newly spawned processes) to it.

hyperthunk:
Ok let me compare them two approaches now. Sorry for the delay!

qnikst:
Closing in favor of #188. Reason: another approach is simpler and give required guarantees, the only issue connected with n-t-backed implementation, and should not be workarounded on d-p level.

PR #188, opened 2015-03-07
A closeLocalNode that cleans up after itself

mboes:
`closeLocalNode` currently closes the network-transport endpoint, but that's about it. Processes that run on the given node still run, it's just that they can no longer be contacted via a network-transport connection. In this PR, we take a page from network-transport-tcp's book and define a node as having two possible states: "valid" and "closed". A node in the "closed" state is defunct, so all resources associated with it can be freed. Moreover, no new processes can be spawned on a closed node without raising an exception.

This PR is an alternative to #187.

Ping @qnikst you said you had tests for this?

Ping @facundominguez

### DIFF
--- a/src/Control/Distributed/Process/Internal/Messaging.hs
+++ b/src/Control/Distributed/Process/Internal/Messaging.hs
@@ -114,7 +114,7 @@ setupConnBetween node from to implicitReconnect = do
           Left _ ->
             return Nothing
           Right () -> do
-            modifyValidLocalState_ (localState node) $
+            modifyValidLocalState_ node $
               return . (localConnectionBetween from to ^= Just (conn, implicitReconnect))
             return $ Just conn
       Left _ ->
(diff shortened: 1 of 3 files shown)


mboes:
> If `closeLocalNode` returns before all processes die, there is no way to know if the transport is being used. This in turn, implies that the transport needs to be made crash free after the endpoint is released

@facundominguez network-transport backends shouldn't just cause the program to segfault if someone attempts to use the transport after it has been closed. network-transport-tcp is safe in this way, using a similar mechanism as in this PR. It's a bug in any network-transport backend that behaves in this way if not, that I don't think it's distributed-process's business to work around. And as noted above, it would be unfortunate to have a call to `closeLocalNode` hang just because some process messed up its exception handlers.

Opening local nodes is very common for tests. It should be possible to close a local node after a test completes and have all resources allocated during that test (namely processes) reclaimed. This kind of resource reclamation is likely why there was a TODO item associated to this function in the first place, calling for killing the processes. Furthermore, processes become pretty useless once the endpoint is closed: they can't communicate anymore with anyone else (technically they can with other local processes, but that's an implementation detail due to an optimization).

hyperthunk:
Reviewing now, so sorry for the delay guys.

qnikst:
I'd like to squash and merge this, any objections?

facundominguez:
LGTM

qnikst:
Ops.. forgot to squash..

PR #190, opened 2015-05-21
Fix queue size counter.

qnikst:
Queue size counter didn't respect messages read from channels
as a result it was possible to have negative values, and
counter showed incorrect size. This commit fixes situation by
adding a "label" to data and decrementing counter only if it
was read from the mailbox.

Fixes: DP-111

### DIFF
--- a/src/Control/Distributed/Process/Internal/CQueue.hs
+++ b/src/Control/Distributed/Process/Internal/CQueue.hs
@@ -10,7 +10,7 @@ module Control.Distributed.Process.Internal.CQueue
   , enqueueSTM
   , dequeue
   , mkWeakCQueue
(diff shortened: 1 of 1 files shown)


qnikst:
@facundominguez please check

facundominguez:
LGTM.

Minor comments:
* `checkArrived :: [m -> Maybe a] -> StrictList m -> (StrictList m, Maybe (Either a a))` never returns `Just (Left ...)` so its changes look unnecessary.
* The changes to `returnOld` are boring. It is probably best to inline and remove the function.

qnikst:
I have fixed check arrived, I agree with inlining functions but as a separate PR after this one will be merged.

PR #191, opened 2015-05-29
Fix say in case of remote loggers.

qnikst:
Currently nsend doesn't work for the hosts that are remote,
so we add this workaround that allowes to send logs to remote host.
This commit should be rolled back once registry will be changed
approrietary.


qnikst:
Closing in favour of #192

PR #193, opened 2015-06-11
Add ghc 7.10 to travis tests

qnikst:


qnikst:
This commit fixes Travis reports.

facundominguez:
This is not running any tests. It just builds them, it seems.

qnikst:
Mimics the previous behaviour (fixed).

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: haskell
 
 ghc:
+ - 7.10
```
mboes: Are you sure this works? I don't know that Travis has 7.10 support yet. Further, I'm pretty sure Travis parses the above the same as 7.1, so the version numbers should really be quoted.

qnikst: Thanks for the pointer, Travis has 7.10 support but it should be written as "7.10", otherwise it will be parsed as 7.1, I'll fix that.

PR #195, opened 2015-06-12
Make 'sendCtrlMsg' strict in the local case.

facundominguez:
Now it evaluates the input message to HNF. If left non-strict, the NC would stop when evaluating a message fails or blocks.

It is still possible to crash the NC if deeper evaluation of the message produces errors. For this sake we make a precondition of 'sendCtrlMsg' that further evaluation of the message should not produce errors in the local case.


mboes:
test case?

facundominguez:
Will come in separate PR. There are no tests in d-p itself.

mboes:
true.

PR #200, opened 2015-06-17
Optimization: do not use network transport for local spawnAsync.

qnikst:


src/Control/Distributed/Process/Internal/Primitives.hs
```diff
@@ -982,7 +983,10 @@ expectTimeout n = receiveTimeout n [match return]
 spawnAsync :: NodeId -> Closure (Process ()) -> Process SpawnRef
 spawnAsync nid proc = do
   spawnRef <- getSpawnRef
-  sendCtrlMsg (Just nid) $ Spawn proc spawnRef
+  node     <- getSelfNode
+  if nid == node
+    then sendCtrlMsg Nothing $! Spawn (DS.force proc) spawnRef
```
facundominguez: `$!` is redundant now that `sendCtrlMsg` is strict in the local case.

facundominguez:
Please, edit the commit message to make explicit the addition of the NFData instances, or better, split in  two commits.

src/Control/Distributed/Process/Internal/Primitives.hs
mboes: Why special case in this function? Surely the same reasoning applies everywhere else? Can't we push this logic to `sendCtrlMsg`? Further, we *must* be consistent in how we obtain semantic consistency between the remote and the local case: currently for better or worse d-p does not use NFData for this purpose but instead `seq (encode x) x`.

In any case `NFData` is overkill here: `Spawn` is strict in its first argument, and `Closure` doesn't normally embed thunks.

mboes:
I'm a -1 on this change in its current form, for the reasons stated above.

Side note: deepseq is a blunt tool, with a performance cost. Use it sparingly.

src/Control/Distributed/Process/Internal/Primitives.hs
facundominguez: > Can't we push this logic to sendCtrlMsg?

How?

> In any case NFData is overkill here: Spawn is strict in its first argument, and Closure doesn't normally embed thunks.

That is not true in our projects. I found at least one case where deep evaluation of the closure produced an error, and it was an expensive error to find.

facundominguez: I'm fine with the `seq (encode x) x` hack.

qnikst: I'm not sure I understood why `seq (encode x) x` is better or cheap here..
Can we have some outline here?

1. should I strictness to `sendCtrlMsg` itself (in `Nothing` branch?
2. should I use `seq (encode x) x` in order to force value in every place where we are sending (or should send but don't do it due to optimization)?
3. should I introduce changes to any other use of `sendCrtlMsg` in this PR?

mboes: > I found at least one case where deep evaluation of the closure produced an error, and it was an expensive error to find as it crashes the NC.

That's a problem with the definition of `Closure`, not something we should be papering over by forcing evaluation defensively everywhere in the code.

facundominguez: I propose we don't involve this change with broader concerns. This is a specific contribution which doesn't need further agreement on the future of `sendCtrlMsg`. The other alternative is to drop this PR entirely as the optimization is not essential for any purpose. It is a subproduct of fixing the d-p stack.

If we were to merge it:

> 1. should I move forcing value procedure to sendCtrlMsg itself (in Nothing branch)?

No. Keep it as it is and discuss further later.

> 2. should I use seq (encode x) x in order to force value in every place where we are sending (or should send but don't do it due to optimization)?

Do it only here. If there are other places where it is needed let us discuss it later.

> 3. should I introduce changes to any other use of sendCrtlMsg in this PR?

No, better to keep that for later.

facundominguez: > That's a problem with the definition of Closure, not something we should be papering over by forcing evaluation defensively everywhere in the code.

Makes sense to me.

qnikst:
Some more notes:

1. we *can't* change `sendCtrl` message because Unsafe Primitives *require* arguments to be unevaluated, so the best we can do, manually check that all safe uses are ok

2. `seq (encode x) x` *do not* guarantee evaluation because only evaluation of a part of a structure that belongs to the 1st chunk of Lazy Bytestring. It may not be a problem for small messages, but it's not true for large messages.

3. `rnf` definitely have less performance cost than using `encode` (less actions, less allocations).

4. I appreciate that for uniformness we may want to use `encode` still.

5. for internal structures (like Closure here) we may want `NF` to the same as `WHNF` then using `seq` or `$!` will be sufficient

qnikst:
I have removed all debatable changes, but it order for this to work reliably
[distributed-static PR](https://github.com/haskell-distributed/distributed-static/pull/6) should be merged first.
@facundominguez @m PTAL.

src/Control/Distributed/Process/Internal/Primitives.hs
```diff
@@ -986,7 +986,10 @@ expectTimeout n = receiveTimeout n [match return]
 spawnAsync :: NodeId -> Closure (Process ()) -> Process SpawnRef
 spawnAsync nid proc = do
   spawnRef <- getSpawnRef
-  sendCtrlMsg (Just nid) $ Spawn proc spawnRef
+  node     <- getSelfNode
+  if nid == node
+    then sendCtrlMsg Nothing $! Spawn proc spawnRef
```
facundominguez: Redundant `$!`.

qnikst: why? w/o `$!` we will have `sendCtrlMsg Nothing  (lazy (Spawn proc spawnRef))`?

facundominguez: check again the latest definition of sendCtrlMsg:
https://github.com/tweag/distributed-process/blob/optimization/spawnAsync/src/Control/Distributed/Process/Internal/Messaging.hs#L195

qnikst:
Fixed

facundominguez:
:+1: 

Change the commit message to something like:

    Have `spawnAsync` not use the transport in the local case.

qnikst:
Done

PR #237, opened 2015-06-18
Fix wrong NFData instance for Message.

### LINKED ISSUE #196: Wrong NFData instance of Message
The NFData instance of Message does not evaluate the message to NF when it is in unencoded form.

It is unclear if this can cause troubles. This instance seems to be used in d-p-extras.

qnikst:
Fixes #196


facundominguez:
LGTM

PR #242, opened 2015-06-25
Add Functor instance for Match data type

Shimuuar:
Main use of that instance is to allow composition of `Match`es of different type:

```.haskell
receiveWait
  [ Left <$> (match :: Match a)
  , Right <$> (match :: Match b)
  ]
```

It's not always convenient/possible to set types of Match right at construction time and I don't see any compelling reason to not add this instance


facundominguez:
Looks good to me. Doesn't `deriving Functor` work for these types?

Shimuuar:
It does. DeriveFunctor extension wasn't enabled so I wrote them manually. Should I update pull request?

facundominguez:
That would be helpful, thx.

Shimuuar:
Done

PR #249, opened 2015-07-15
Don't hold the localState MVar while calling potentially blocking operations.

facundominguez:
The node controller could be blocked if the MVar is not promptly refilled.

### DIFF
--- a/src/Control/Distributed/Process/Internal/Messaging.hs
+++ b/src/Control/Distributed/Process/Internal/Messaging.hs
@@ -15,7 +15,9 @@ import qualified Data.ByteString.Lazy as BSL (toChunks)
 import qualified Data.ByteString as BSS (ByteString)
 import Control.Distributed.Process.Serializable ()
 
+import Control.Concurrent (forkIO)
 import Control.Concurrent.Chan (writeChan)
+import Control.Exception (mask_)
 import Control.Monad (unless)
 import Control.Monad.IO.Class (liftIO)
 import Control.Monad.Reader (ask)
@@ -30,6 +32,7 @@ import qualified Network.Transport as NT
 import Control.Distributed.Process.Internal.Types
   ( LocalNode(localState, localEndPoint, localCtrlChan)
   , withValidLocalState
+  , modifyValidLocalState
   , modifyValidLocalState_
   , Identifier
   , localConnections
@@ -51,6 +54,7 @@ import Control.Distributed.Process.Internal.Types
   , Identifier(NodeIdentifier, ProcessIdentifier, SendPortIdentifier)
   )
 import Control.Distributed.Process.Serializable (Serializable)
+import Data.Foldable (forM_)
 
 --------------------------------------------------------------------------------
 -- Message sending                                                            --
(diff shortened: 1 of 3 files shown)


src/Control/Distributed/Process/Internal/Messaging.hs
```diff
@@ -135,24 +137,28 @@ connBetween node from to implicitReconnect = do
         setupConnBetween node from to implicitReconnect
 
 disconnect :: LocalNode -> Identifier -> Identifier -> IO ()
-disconnect node from to =
-  modifyValidLocalState_ node $ \vst ->
+disconnect node from to = do
```
qnikst: don't we need `mask_` here?

qnikst:
Looks great! modulo question about `mask_`

src/Control/Distributed/Process/Internal/Messaging.hs
facundominguez: We want the connection to be closed eventually, but masking assumes that NT.close is uninterruptible or it won't accomplish anything. And I don't think NT.close is uninterruptible.

Maybe mask and close the connection with `forkIO $ forM_ mio id`?

qnikst: > We want the connection to be closed eventually, but masking assumes that NT.close is uninterruptible or it won't accomplish anything. And I don't think NT.close is uninterruptible.

I'd like n-t semantics to say smth here, however I agree with the point.
I think `mask_ $ do { ... ; forkIO $ forM_ mio id }`, like you are suggesting will be a sane solution.

facundominguez:
@mboes, maybe you should take a look. This is another chapter in stopping synchronous transport operations from blocking the NC.

src/Control/Distributed/Process/Node.hs
```diff
@@ -389,7 +390,9 @@ forkProcess node proc =
                 (return . DiedException . (show :: SomeException -> String)))]
 
           -- [Unified: Table 4, rules termination and exiting]
-          modifyValidLocalState_ node (cleanupProcess pid)
+          mconns <- modifyValidLocalState node (cleanupProcess pid)
+          forM_ mconns (mapM_ NT.close)
```
mboes: You're saying `forkIO` [here](https://github.com/haskell-distributed/distributed-process/pull/249/files#diff-feba37ab068c8749d02e4a8c30d18d41R163), whereas you're not here. Why is that? I would argue the NT.close should always be one-sided and asynchronous, hence no need to `forkIO`. But, the current API makes it difficult indeed for transports to guarantee that, so maybe `forkIO` consistently here but add an XXX with a reference to the ticket about making the network-transport API asynchronous?

mboes:
:+1: modulo my comment above.

src/Control/Distributed/Process/Node.hs
facundominguez: `forkIO` was meant to prevent the clean up from being interrupted in other places. In this occurrence, the clean up happens in a scope where exceptions are uninterruptibly masked, thus forkIO is not needed for that sake.

I do see that notifying the NC of the process death could delay. We could move closing connections after that, or use `forkIO` for uniformity as you are considering.

There is also the chance of threads trying to deliver exceptions to this one to block until the cleanup finishes. Another reason to use `forkIO`.

mboes: Fine by me either way.

Issue #55, opened 2012-10-23
[DP-11] unmonitor should remove monitor messages from the mailbox (if any)


edsko:
[Imported from JIRA. Reported by Edsko de Vries @edsko)  as [DP-11](https://cloud-haskell.atlassian.net/browse/DP-11) on 2012-10-23 11:23:33]

facundominguez:
There is a PR addressing this issue, and I'm wondering what the purpose of removing the notifications from the mailbox is. Notifications can arrive anytime after unmonitoring, so I don't see how this helps avoiding a message leak.

VicNumber21:
May you please provide a use case when the caller of unmonitor may be interested in handling monitor notification?
My understanding is that unmonitor is called when you are not interested in the monitored resource anymore? Am I missing something?

facundominguez:
I have no use case. Just pointing at the fact for others to consider.

hyperthunk:
I think part of the rationale for this is provided the same behaviour as Erlang/OTP. In that case, the `unmonitor` call flushes (or there's a flushing variant, I forget which is the case) so you won't get unexpected messages in the mailbox. This is particularly important in stateless gen_server and gen_fsm implementations (e.g., what we do in distributed-process-client-server), where an unexpected monitor signal might trigger a shutdown or restart that we don't actually want, and the server loop and message handling/receiving code are hidden from the caller.

facundominguez:
In that case, I would find less surprising to offer a variant that does the flushing of monitor notifications. Erlang provides more than one variant at least.

hyperthunk:
Yes @facundominguez that's right, and I agree we should offer a variant rather than changing the current behaviour.

mboes:
Erlang does not flush by default, but as @hyperthunk says, gives you the option to. Makes sense to copy Erlang, with an `unmonitorFlush` variant or somesuch.

mboes:
Well, except that Erlang only kept the non-flush variant by default due to backwards compatibility... http://erlang.org/pipermail/erlang-questions/2008-August/037564.html

hyperthunk:
[...]

PR #252, opened 2015-09-08
DP-11: Remove monitor messages in unmonitor

VicNumber21:


facundominguez:
Thanks, Victor. Which problem are you trying to address with this patch? I'm making a guess, but I don't think the patch solves it.

VicNumber21:
@facundominguez you're welcome.

I've tried to remove messages from message queue to prevent catching them after unmonitor called.
I found the task description here https://cloud-haskell.atlassian.net/projects/DP/issues/DP-11?filter=allopenissues and decided to try to address it.

May you please explain why it is not solution for the problem? I'll rework if required.

facundominguez:
We have moved tickets from Jira to github. Here is the github one:
https://github.com/haskell-distributed/distributed-process/issues/55

I'm not sure I make sense of the problem. Will discuss it there.

VicNumber21:
I see. Probably it is required to update http://haskell-distributed.github.io/ (Issues link at the bottom) and http://haskell-distributed.github.io/wiki/contributing.html (point 1, link to Jira there) to remove reference to Jira.
It may help to newcomers to get things right.

I'll track the discussion in the issue and wait for your decision.

mboes:
@facundominguez any closure on #55 and what to do with this PR?

src/Control/Distributed/Process/Internal/Primitives.hs
```diff
@@ -917,9 +919,17 @@ unlinkPort sport = do
 unmonitor :: MonitorRef -> Process ()
 unmonitor ref = do
   unmonitorAsync ref
-  receiveWait [ matchIf (\(DidUnmonitor ref') -> ref' == ref)
-                        (\_ -> return ())
-              ]
+  receiveWait [ matchIf isDidUnmonitor ignore ]
+  receiveAllNow [ matchIf isProcessMonitorNotification ignore,
```
facundominguez: I guess there is no need to collect a list of units `[()]` as receiveAllNow was defined to do below. How about writing this in a single loop here:
```
fix $ \loop ->
  m <- receiveTimeout 0 [...]
  forM_ m loop
```

src/Control/Distributed/Process/Internal/Primitives.hs
```diff
@@ -917,9 +919,17 @@ unlinkPort sport = do
...
+                  matchIf isNodeMonitorNotification ignore,
+                  matchIf isPortMonitorNotification ignore
+                ]
+  return ()
+    where ignore _ = return ()
+          isDidUnmonitor (DidUnmonitor ref') = ref' == ref
+          isProcessMonitorNotification (ProcessMonitorNotification ref' _ _) = ref' == ref
+          isNodeMonitorNotification (NodeMonitorNotification ref' _ _) = ref' == ref
+          isPortMonitorNotification (PortMonitorNotification ref' _ _) = ref' == ref
```
facundominguez: Please, inline all of these local definitions.

src/Control/Distributed/Process/Internal/Primitives.hs
```diff
@@ -1003,6 +1013,25 @@ catchesHandler handlers e = foldr tryHandler (throw e) handlers
 expectTimeout :: forall a. Serializable a => Int -> Process (Maybe a)
 expectTimeout n = receiveTimeout n [match return]
 
+-- | Synonim of 'receiveTimeout' 0.
+--
+-- Immediatelly take a matched message from the message queue.
+receiveNow :: [Match b] -> Process (Maybe b)
+receiveNow = receiveTimeout 0
```
facundominguez: Please, inline this at the (only) call site.

src/Control/Distributed/Process/Internal/Primitives.hs
```diff
@@ -917,9 +919,17 @@ unlinkPort sport = do
 unmonitor :: MonitorRef -> Process ()
```
facundominguez: Please, update the documentation to explain how the mailbox is affected.

facundominguez:
@VicNumber21, I just did a review pass making some observations. Also, could you add a mention of #55 in the commit message?

src/Control/Distributed/Process/Internal/Primitives.hs
facundominguez: Actually, it looks to me like there can be at most one monitor notification in the mailbox for a given `MonitorRef`. Therefore no loop would be necessary. It should suffice:
```Haskell
_ <- receiveTimeout 0 [ ... ]
```

facundominguez:
Would be merging as:
https://github.com/haskell-distributed/distributed-process/pull/268

PR #254, opened 2015-10-13
Avoid racy blocking behavior in closeLocalNode.

facundominguez:
When a process dies, it tries to acquire the localState mvar while masking exceptions uninterruptibly. In turn, closeLocalNode tries to kill processes while holding the localState mvar, thus causing a deadlock.

This patch kills processes only after refilling the mvar.


mboes:
:+1:

PR #255, opened 2015-10-15
Suppress console messages complaining about closed nodes.

facundominguez:
The handler of incoming events and the node controller could both print unhandled exceptions when closing the node.

This patch defines an exception type NodeClosedException for when primitives notice that the node has been closed, and installs a top-level handler that suppresses messages to the console for this
exception.


hyperthunk:
+1 for merging this.

mboes:
:+1:

PR #257, opened 2015-10-18
Make CI build package.

qnikst:
Distributed-process package contains no test suites. As a result
cabal test command does nothing. This commit run cabal build, so
CI will check at least build failures.


qnikst:
This PR is supposed to fail because current master is unbuildable state.

mboes:
Merging now because this fixes CI. But @facundominguez we'll need a better solution down the road.

PR #258, opened 2015-10-19
Move CI to travis containers and build with ghc 7.10.2.

facundominguez:


facundominguez:
cc @mboes

facundominguez:
binutils-dev is not installed, don't know why yet. Anyone knows what is that needed for? It was in the former .yml file?

mboes:
Do we need binutils-dev? Maybe we don't anymore. Don't know why it was there to begin with.

:+1:

qnikst:
LGTM

PR #259, opened 2015-10-19
Reimport distributed-process-tests.

mboes:
The code in d-p-tests used to be local to d-p, not a separate package.
We split this code off into a separate package for the same reason
that n-t-tests is its own package: to allow network-transport backends
to use the d-p-tests to check that they are compatible with what
distributed-process expects.

However, keeping d-p-tests separate has been a pain for CI. And it
makes little sense to keep the histories of the packages separate,
since the two are tightly coupled anyways. So what this patch does is
reimport d-p-tests into the repo, but keep is a separate package. This
poses exactly zero challenges to `stack` users: the provided
`stack.yaml` makes it so that `stack build --test` works
out-of-the-box. `cabal-install` users need to `add-source` in
a sandbox, as we already did in `.travis.yml`.

As part of this change, we no longer build against development
versions of dependencies. We assume that dependencies have their own
release lifecycle, so we should really be testing against the release
versions. If they didn't, then we'd be far better off with a single
repo for all of Cloud Haskell (the way it used to be...).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,10 +23,10 @@ before_install:
   - sed -i 's/^jobs:/-- jobs:/' $HOME/.cabal/config
   - cabal sandbox init
   - cabal sandbox add-source .
-  - cabal sandbox add-source distributed-process-tests
+  - cabal sandbox add-source distributed-process-tests/
 
 install:
-  - cabal install --only-dependencies --jobs=2
+  - cabal install --only-dependencies distributed-process-tests/
 
 script:
-  - cabal install --run-tests
+  - cabal install --run-tests distributed-process-tests/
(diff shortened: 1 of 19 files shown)


facundominguez:
@mboes, I cherry-picked your commits and forced pushed. Your old branch is now `import-d-p-tests-mb`.

mboes:
CI fails but for reasons unrelated to this patch series. d-p-tests is just too flaky.

facundominguez:
In that case, should we refrain temporarily from running the tests during CI?

hyperthunk:
I definitely vote for running some tests! :)

By all means let's temporarily disable those which are proving unreliable until we can improve them.

facundominguez:
LGTM

facundominguez:
Rebased master and resolved conflicts.

facundominguez:
Hm, there are more failing tests. We will have to group them under a category of flaky tests and skip it in ci.

PR #260, opened 2015-10-26
Fix race in runProcess when it fails.

facundominguez:
runProcess would yield control back to the caller and then it would try to kill it. This could cause some strange behavior in tests when they fail.

Moreover, the former solution assumed exceptions were not masked uninterruptibly. This patch has runProcess work even in such cases.

cc: @hyperthunk


mboes:
:+1: nice catch.

hyperthunk:
Looks good to me, and yes, nice catch!

PR #262, opened 2015-11-27
Have nsendRemote skip the transport for local communication.

facundominguez:
This is useful when doing a loop like:
```Haskell
nbroadcast nids label msg = forM_ nids $ \nid -> nsendRemote nid label msg
```
so one doesn't have to treat specially the local case.


qnikst:
:+1:

PR #264, opened 2015-12-01
Stop using the transport for local communication.

facundominguez:
The NC was still using the transport to communicate with itself or with local processes when using the primitives in C.D.P.I.Messaging.

As a side effect, this patch also stops creating connections involving processes when the NC sends a message. Only the node-to-node connection is used by the NC to send messages.

### DIFF
--- a/src/Control/Distributed/Process/Node.hs
+++ b/src/Control/Distributed/Process/Node.hs
@@ -169,6 +169,7 @@ import Control.Distributed.Process.Internal.Types
   , payloadToMessage
   , messageToPayload
   , createUnencodedMessage
+  , unsafeCreateUnencodedMessage
   , runLocalProcess
   , firstNonReservedProcessId
   , ImplicitReconnect(WithImplicitReconnect,NoImplicitReconnect)
@@ -198,7 +199,6 @@ import Control.Distributed.Process.Management.Internal.Types
 import Control.Distributed.Process.Serializable (Serializable)
 import Control.Distributed.Process.Internal.Messaging
   ( sendBinary
-  , sendMessage
   , sendPayload
   , closeImplicitReconnections
   , impliesDeathOf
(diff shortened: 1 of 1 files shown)


facundominguez:
@hyperthunk, you may want to check this in case I'm affecting intended semantics of any ops.

hyperthunk:
The only problem I have with this is that it breaks tracing on the local node. If we can fix that, then I don't think it changes the semantics

facundominguez:
Thanks for taking a look. How is tracing broken?

src/Control/Distributed/Process/Node.hs
```diff
@@ -661,6 +661,30 @@ instance Show ProcessKillException where
   show (ProcessKillException pid reason) =
     "killed-by=" ++ show pid ++ ",reason=" ++ reason
 
+ncSendToProcess :: ProcessId -> Message -> NC ()
+ncSendToProcess pid msg = do
+    node <- ask
+    if processNodeId pid == localNodeId node
+      then ncEffectLocalSend node pid msg
```
facundominguez: hm, this line https://github.com/haskell-distributed/distributed-process/pull/264/files#diff-8684dec0b32dae52a61ac233d229266cL521
should be move here.

facundominguez:
Let me try to explain how this is broken.

When a message is traced, the tracer sends a `NameSend "trace.logger" ...` message to the NC. The NC calls `ncEffectLocalSend` when dealing with this, which calls `trace` which sends again a `NameSend "trace.logger" ...` message and so on.

I checked master and it looks like it was broken in this sense already. I believe this patch broke it
https://github.com/haskell-distributed/distributed-process/commit/ced6592890eacfb4e35303757bacf98850f86f6f

@qnikst ?

Do we agree that this PR is not worsening tracing then?

qnikst:
When I reverted that commit I was not thinking about tracing unfortunately and nobody pointed that out to me. I was thinking about fixing behaviour we rely on in simple-localnet. 
I'm ok with not worsening tracking now, but we should open an issue about it then.

hyperthunk:
Yes please, tracing is vital to debugging message passing applications. If you open an issue, I'll self assign and take a look at fixing it.

hyperthunk:
Argh, sorry I didn't review this properly. We now have this logic in two places... the `send` implementation in Internal/Primatives.hs already skips the transport in the local case... Erm... ?

facundominguez:
Ok. Though I wouldn't know how to reuse that logic in the NC.

hyperthunk:
I'm sure we don't need to. For a local trace, that bypasses the nc altogether, we need the tracing code at the call site. It made sense (when I originally implemented this) to carry the tracer refs in the local node state and access it at the call site in both cases, for send, and to deal with receive in the nc code. And there is a certain symmetry to that approach. But if we want to trace sends in the nc, we will have to live with both code blocks, as long as we don't trace twice (or worse, not at all), I'm fine with either approach.

PR #266, opened 2016-01-06
Throw NodeClosedException when the node controller is stopped.

facundominguez:
Formerly the NC could end with either `NodeClosedException` or with `ThreadKilled`. ThreadKilled was a bit confusing because there is no call to `threadKill` to send it and there was no good reason to prefer it over `NodeClosedException`.


qnikst:
LGTM

hyperthunk:
+1

Issue #265, opened 2015-12-02
Tracing is broken


hyperthunk:
Tracing for named send, unreliable send, and monitoring appears to be either missing or has been stripped out accidentally in previous commits. I'll look at putting this back in.

facundominguez:
Not sure if it is meant to be fixed here.

Tracing received messages currently produces a loop, which I hope to fix [here](https://github.com/haskell-distributed/distributed-process/commit/bf0bea796c5beda81a6347fe86663a7585d135f7).
It is not very elegant but it should work while something better is implemented.

I will make a PR after I write some tests.

hyperthunk:
That code makes sense, as does the bug. Silly mistake on my part.. Thanks for the spot, and fix!

hyperthunk:
I'm just starting to cook up a PR for this..

hyperthunk:
I am working on proving the things that we're saying in #241

PR #267, opened 2016-01-22
Avoid loop when tracing received messages.

facundominguez:
When a trace message was sent to "trace.logger", the NC would produce a trace message to indicate the arrival of it, this new trace message would in turn have the NC produce another trace message and so on.

In this patch the NC does not produce a trace message when the destination is the "trace.logger" process.

Test will be in an upcoming PR of distributed-process-tests.

### DIFF
--- a/distributed-process-tests/src/Control/Distributed/Process/Tests/Tracing.hs
+++ b/distributed-process-tests/src/Control/Distributed/Process/Tests/Tracing.hs
@@ -5,6 +5,7 @@ module Control.Distributed.Process.Tests.Tracing (tests) where
 import Control.Distributed.Process.Tests.Internal.Utils
 import Network.Transport.Test (TestTransport(..))
 
+import Control.Applicative ((<*))
 import Control.Concurrent (threadDelay)
 import Control.Concurrent.MVar
   ( MVar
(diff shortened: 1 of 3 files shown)


facundominguez:
hm, had forgotten that tests live now on the same repo as d-p. I just included the tests in this PR then.

facundominguez:
Related to #265.

distributed-process-tests/src/Control/Distributed/Process/Tests/Tracing.hs
```diff
@@ -19,9 +20,13 @@ import Control.Distributed.Process.Debug
 import Control.Distributed.Process.Management
   ( MxEvent(..)
   )
+import qualified Control.Exception as IO (bracket)
+import Data.List (isPrefixOf, isSuffixOf)
 
 #if ! MIN_VERSION_base(4,6,0)
 import Prelude hiding (catch, log)
+#else
+import Prelude hiding ((<*))
```
qnikst: Latest prelude that exported `catch` in prelude 4.4.1, so I suppose we could drop this CPP thing and leave
only `import Prelude` (in case `log` should not be hidden)

qnikst:
Fails on 7.4 (can be removed from travis) and 7.6 due to build error, `System.Environment`didn't have `setEnv` and `unsetEnv` methods.

qnikst:
Everything else looks good to me.

hyperthunk:
We handle those issues in d-process with conditional compilation, can't we do the same here? I mean, distributed-process uses {set,get}Env, in the tracing code...

facundominguez:
> We handle those issues in d-process with conditional compilation.

Well, issues are solved now.

PR #274, opened 2016-02-24
Implement MonadCatch, MonadThrow, MonadMask for Process.

qnikst:
Implements monad catch family for process monad, this heavily simplifies implementation of the different exception handling methods and simplify writing more general code.

### DIFF
--- a/src/Control/Distributed/Process/Internal/Types.hs
+++ b/src/Control/Distributed/Process/Internal/Types.hs
@@ -4,6 +4,9 @@
 {-# LANGUAGE GADTs  #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE Rank2Types #-}
 
 -- | Types used throughout the Cloud Haskell framework
 --
@@ -112,11 +115,12 @@ import Control.Concurrent (ThreadId)
 import Control.Concurrent.Chan (Chan)
 import Control.Concurrent.STM (STM)
 import Control.Concurrent.STM.TChan (TChan)
+import Control.Monad.Catch (MonadThrow(..), MonadCatch(..), MonadMask(..))
 import qualified Network.Transport as NT (EndPoint, EndPointAddress, Connection)
-import Control.Applicative (Applicative, Alternative, (<$>), (<*>))
+import Control.Applicative
 import Control.Monad.Fix (MonadFix)
 import Control.Monad.Reader (MonadReader(..), ReaderT, runReaderT)
-import Control.Monad.IO.Class (MonadIO)
+import Control.Monad.IO.Class (MonadIO(..))
 import Control.Distributed.Process.Serializable
   ( Fingerprint
   , Serializable
@@ -139,6 +143,7 @@ import qualified Control.Distributed.Process.Internal.StrictContainerAccessors a
 
 import Data.Hashable
 import GHC.Generics
+import Prelude
 
 --------------------------------------------------------------------------------
 -- Node and process identifiers                                               --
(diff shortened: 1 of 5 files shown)


hyperthunk:
What effect does this have on code using the existing primitives? And are the semantics the same? They're used extensively in the platform libraries.. 

Also do we need to document these changes? And which release will this be earmarked for? 0.7?

qnikst:
This change doesn't introduce any change in semantics or how primitives works (at least currently). So all the code that was correct before this patch is still correct after this patch and will have exactly the same semantics.
As an additional benefit it will allow to use primitives from other libraries (e.g. `exceptions`) for `Process` monad. For example: `withLock` from d-p extras can be implemented as:
```
+withLock :: (Exclusive e, MonadMask m, MonadIO m) => e -> m a -> m a
+withLock excl = bracket_ (liftIO $ acquire excl)
+                         (liftIO $ release excl)
```
this will generalize both `withLock` and `withLockP` and will not have several races that exists in current code in presence of exceptions.

I think we need to add changes to changelog only, this code should go in next major release, i.e. 0.7, (I don't expect that to happen really soon).

hyperthunk:
Ok this sounds like an ideal solution, so +1 from me. Would you mind logging an issue in d-p-extras so I remember to update that code - it's used in a lot of test cases for platform libs and could explain why we see intermittent failures on travis.

I want to start maintaining the platform libs again, get them back into ci and make sure everything gets tested and bundled by the time the next major release goes out, so I'll take ownership of any issues where we need to modify platform code to take advantage of this.

Also, would you mind pointing out to me the concurrency bugs in the existing implementations at some point please, as the better I understand these things, the better the quality of my contributions to the codebase will be in future? Thanks!!

qnikst:
About testing, we have stack based tests in cloud-haskell package, that tests all of the packages, though it seems we will need to write some hooks so on PR to one of the individual packages tests in cloud-haskell will run. @mboes do you have any insights here, it seems you have implemented similar hooks for our internal project?

About concurrency bug, of cause I'll notify you if I'll see any, I planned to send a PR after this one will land.

hyperthunk:
Maybe we should break out the conversation about testing so as not to derail this PR and raise an issue in d-p-tests? I'll do that now...

src/Control/Distributed/Process/Internal/Types.hs
```diff
@@ -345,7 +350,37 @@ data LocalProcessState = LocalProcessState
 newtype Process a = Process {
     unProcess :: ReaderT LocalProcess IO a
   }
-  deriving (Functor, Monad, MonadIO, MonadReader LocalProcess, Typeable, Applicative, MonadFix)
+  deriving ( Applicative, Functor, Monad, MonadFix, MonadIO
```
mboes: minor style point: if you're going to introduce a line break here, one class name per line will make it more manageable.

mboes:
Looks good. However, as it stands, this is a breaking API change: `catch`, `mask` and friends are no longer exported from `Control.Distributed.Process`. Arguably, that's as it should be, since reexports are pretty evil (we'd have to reexport a `catch` that doesn't mention `Process` at all), but it's easy enough to introduce backwards compatibility here, so let's do this:

* keep wrappers in `Control.Distributed.Process`.
* Mark them as deprecated.
* Remove them in a few releases.

hyperthunk:
+1 for the approach @mboes suggests.

qnikst:
@mboes @hyperthunk please take another look, I've addressed comments. By default I'm exporting specialized versions of the functions marked as being deprecated, this will lead to a number of warnings in downstream packages, but I suppose that it will point out places that need closer look so is fine.

qnikst:
Also I fixed compatibility warnings in the modules where I have changed imports, I hope it's fine to not extract those changes in a separate commit.

hyperthunk:
I'll set to work updating the platform libraries.

PR #268, opened 2016-01-29
Remove monitor message in unmonitor.

facundominguez:
Thanks to Victor Portnov.

Addresses #55.

@hyperthunk, could you take a look?

This patch makes `TestMx` fail:
```
    Mx Agents:
      Event Handling: [OK]
      Inter-Agent Broadcast: [OK]
      Agent Mailbox Handling: [OK]
      Agent Dual Input Handling: [OK]
      Agent Input Prioritisation: [OK]
    Mx Global Properties:
      Global Property Publication: [OK]
      Clearing Global Properties: [OK]
      Purging Global Tables: [OK]
      Deleting and (Re)Creating Global Tables: [Failed]
    expected (Just 15, Nothing, Just 15): invalid table entry found!
      Wait: [OK]
    
             Test Cases   Total       
     Passed  9            9           
     Failed  1            1           
     Total   10           10          
```

PR #276, opened 2016-03-02
Remove Mx Data Tables

hyperthunk:
This API isn't used, is easy to replace with various other packages, and contains races that are leading to failing test cases in CI.

This resolves the issue blocking us from merging #268.


facundominguez:
LGTM

PR #282, opened 2016-03-22
With monitor ref

qnikst:
Alternative approach to with monitor ref here we change withMonitor to have monitor ref and are introducing withMonitor_ for old behaviour. If we need to support compatible version - we could fork from d51d1e0.

### DIFF
--- a/src/Control/Distributed/Process.hs
+++ b/src/Control/Distributed/Process.hs
@@ -102,7 +102,7 @@ module Control.Distributed.Process
   , monitorPort
   , unmonitor
   , withMonitor
-  , withMonitorRef
+  , withMonitor_
   , MonitorRef -- opaque
   , ProcessLinkException(..)
   , NodeLinkException(..)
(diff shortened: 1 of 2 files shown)


src/Control/Distributed/Process.hs
```diff
@@ -270,7 +270,11 @@ import Control.Distributed.Process.Internal.Primitives
   , monitorPort
   , unmonitor
   , withMonitor
+<<<<<<< HEAD
   , withMonitorRef
+=======
+  , withMonitor_
+>>>>>>> 25e7569... Prepare for changing withMonitor.
```
facundominguez: noise

facundominguez:
LGTM

mboes:
Looks good, but why is the PR bumping version numbers? Even when making API changes, what we've been doing so far is have only the maintainer bump version numbers, and only when cutting a new release. Any reason to change that?

Also, let's keep `withMonitorRef` and simply deprecate it for now. That's a gratuitous breaking change that we can avoid.

facundominguez:
IIRC, `withMonitorRef` is not part of any release yet.

mboes:
Ah. Ok then. I thought it had made it in the last one.

qnikst:
I've addressed version number increase issue, please take another look.

PR #283, opened 2016-03-29
Adapt to GHC 8.0 interface changes

bgamari:
* `mkWeak#` now expects a raw state transformer instead of a boxed `IO` action.
 * the `template-haskell` `VarI` constructor has lost a field

### DIFF
--- a/distributed-process.cabal
+++ b/distributed-process.cabal
@@ -42,7 +42,7 @@ flag old-locale
 
 Library
   Build-Depends:     base >= 4.4 && < 5,
-                     binary >= 0.6.3 && < 0.8,
+                     binary >= 0.6.3 && < 0.9,
                      hashable >= 1.2.0.5 && < 1.3,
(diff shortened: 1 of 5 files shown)


bgamari:
It may be worthwhile to introduce a `Control.Distributed.Process.Internal.Compat` module to expose a wrapper `mkWeak` to keep the CPP-nastyness isolated in one place. Let me know if you'd prefer this.

mboes:
Looks (nearly) fine as-is IMHO. This patch could do with less code duplication: couldn't you duplicate only the LHS and not the RHS (which includes a number of gnarly lexemes and identifiers that may well diverge over time)?

facundominguez:
LGTM

facundominguez:
Also, thanks for the patch!

bgamari:
@mboes, BTW, the patch has been updated.

PR #285, opened 2016-05-02
Don't throw exceptions asynchronously.

facundominguez:
This undoes the fix in 07dbb9b3 since it breaks unlink semantics. With exceptions sent asynchronously, a ProcessLinkException could be delivered to a process after it has unlinked.

To avoid introducing a deadlock when forkProcess cleans up, we release the node state MVar before calling throwTo.


mboes:
:+1:

PR #287, opened 2016-06-02
Node: Provide more detail in invalidRequest messages

bgamari:
I noticed a few of these errors fly by while debugging. I don't know precisely the reason yet, but it seems like we should offer more information when they occur.

### DIFF
--- a/src/Control/Distributed/Process/Node.hs
+++ b/src/Control/Distributed/Process/Node.hs
@@ -516,7 +516,7 @@ handleIncomingMessages node = go initConnectionState
                     . (incomingFrom theirAddr ^: Set.insert cid)
                     $ st
                     )
-            else invalidRequest cid st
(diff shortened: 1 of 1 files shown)


src/Control/Distributed/Process/Node.hs
```diff
@@ -516,7 +516,7 @@ handleIncomingMessages node = go initConnectionState
                     . (incomingFrom theirAddr ^: Set.insert cid)
                     $ st
                     )
-            else invalidRequest cid st
+            else invalidRequest cid st "unknown reliability"
```
facundominguez: I'd say `"unsupported reliability: " ++ show rel`.

bgamari: Fixed.

facundominguez:
Thanks @bgamari. Took me long to get to review these messages.

Will merge in
https://github.com/haskell-distributed/distributed-process/pull/290

PR #288, opened 2016-06-05
Bump transformers and ghc-prim upper bounds

bgamari:
GHC 8.0.1 ships with these.


facundominguez:
Thanks. I think we can perhaps get rid of the ghc-prim dependency. Will look at that later.

PR #291, opened 2016-07-06
Say: Send unformatted time over the wire

bgamari:
It's not clear why this function was designed as it was but it makes it
very difficult to feed messages from `say` to a proper logging
framework.

Unfortunately, it's not clear how to push out this sort of change 
without silently breaking user code due to the untyped nature of
 message passing. Thoughts?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2016-07-06 Facundo Domínguez <facundo.dominguez@tweag.io> 0.7.0
+
+* Change type of message sent by `say` from a 3-tuple to a proper
+type (`SayMessage`) with a proper `UTCTime`.
+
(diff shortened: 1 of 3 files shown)


facundominguez:
CI is failing.
LGTM otherwise.

Can't think of an approach which warns the user of the breakage without ugly hacks (like having receiveWait fail if the `SayMessages` are in the mailbox while `(String, ProcessId, String)` is expected.).

bgamari:
Fixed; I've also added a changelog entry for a 0.7.0 release since this almost certainly needs a major version bump.

facundominguez:
CI still fails.

bgamari:
But it passes now! ;)

Sorry for the large iteration count.

PR #292, opened 2016-07-06
Fix various warnings

bgamari:
Just as it says on the tin.


src/Control/Distributed/Process/Serializable.hs
```diff
@@ -63,7 +61,7 @@ decodeFingerprint :: ByteString -> Fingerprint
 decodeFingerprint bs
   | BS.length bs /= sizeOfFingerprint =
       throw $ userError "decodeFingerprint: Invalid length"
-  | otherwise = BSI.inlinePerformIO $ do
+  | otherwise = unsafePerformIO $ do
```
facundominguez: What's the motivation for this?

facundominguez: Never mind.

facundominguez:
LGTM

PR #295, opened 2016-08-26
Remove monitors from remote nodes.

facundominguez:
When a process dies, we need to remove it from the monitor lists of any nodes where there is a process being monitored by the dead process.


qnikst:
:+1:

PR #296, opened 2016-08-26
Use only one connection to communicate with NCs.

facundominguez:
This avoids connections from processes to nodes, which are never released for long-running processes. Given the implicit reconnections considered in the semantics for connections to/from node controllers, this change wouldn't alter the expected behavior of primitives.


qnikst:
:+1:

PR #297, opened 2016-08-26
Use usend to implement spawn.

facundominguez:
This avoids the need to create a connection from the spawner to the spawned process when spawning.

The formal semantics are expressed for `spawnAsync` only.


qnikst:
I agree with this change. What makes me think about it is that if it's preserves required semantics, but seems that everything is fine here.

PR #309, opened 2017-02-26
Fix haddock for "try", it refers to "mask_"

lolepezy:
I looks like a copy-paste problem in the haddock comment.


facundominguez:
Thanks @lolepezy.

PR #311, opened 2017-03-08
Use correct Exception Handling in MxAgents

hyperthunk:
Running an agent with a finalizer is intended to ensure the finalizer runs in the presence of exceptions, but it was previously run on when an exception occurs!

### DIFF
--- a/src/Control/Distributed/Process/Management.hs
+++ b/src/Control/Distributed/Process/Management.hs
@@ -280,7 +280,7 @@ import Control.Distributed.Process.Management.Internal.Types
 import Control.Distributed.Process.Serializable (Serializable)
 import Control.Monad.IO.Class (liftIO)
 import Control.Monad.Reader (ask)
-import Control.Monad.Catch (onException)
+import Control.Monad.Catch (finally)
 import qualified Control.Monad.State as ST
   ( get
   , modify
@@ -429,7 +429,7 @@ mxAgentWithFinalize mxId initState handlers dtor = do
              -> Process ()
     runAgent eh hs cs c s =
       runAgentWithFinalizer eh hs cs c s
-        `onException` runAgentFinalizer eh s
+        `finally` runAgentFinalizer eh s
 
     runAgentWithFinalizer :: MxAgent s ()
                           -> [MxSink s]


hyperthunk:
Note: this is part of a series of patches fixing the management/tracing infrastructure. The rest I'm still testing atm.

hyperthunk:
> I do not understand this change, it seems it doesn't change behavior, it seem that it's not possible to exit the loop, rather that by receiving exception or calling error.

Ah, I've actually misread my own code *facepalm* and this change is actually wrong. We call the finalizer if we see the `MxAgentDeactivate` action, an `onException` is there to ensure it is run if we fail. Actually however, this is more nuanced, and requires both masking and catching in the case of `MxAgentDeactivate` to ensure the `onException` block doesn't execute our finalizer twice where (a) an async exception interrupts the finalizer or (b) the finalizer throws a synchronous exception.

> Also (unrelated to the change) it seems that the action is not tail recursive. The onException/finalize is a bracket like call this means that we will chain handlers on each call to runAgent that we in runAgentWithFinalizer

This is not accurate. At least, my understanding of functions defined using `where` is that they can form tail recursive inner bindings, and `runAgentWithFinalizer` is tail recursive with `runAgent`. If I have misunderstood this, then all of the platform libraries are broken (viz managed process), but on review @mboes stated that the `serve` loop for -client-server is indeed tail recursive, and it works exactly the same way (see [here](https://github.com/haskell-distributed/distributed-process-client-server/blob/master/src/Control/Distributed/Process/ManagedProcess/Internal/GenProcess.hs#L347) for example).

I will close and re-work the patch with the relevant mask/catch in place instead.

qnikst:
I'm not about tail recursion but about other issue and the pattern mentioned in another place does not hold. Sorry for a bad notation, but let me explain (`=>` will denote what we are evalating) `{} for branch we appear in)
  1.
```
(runAgent =>)
```
   2.
```
(runAgentWithFinalizer =>) `finally` finalizer
```
   3. 
```
(runAgentWithFinalizer => {MxAgentReady}) `finally` finalizer
```
   4.
```
(runAgent => ) `finally` finalizer
```
    5.
```
( (runAgentWithFinalizer =>) `finally` finalizer) `finally` finalizer
```

So we add additional layer if `runAgentWithFinalizer` calls `runAgent`. This can be resolved by introducing a single toplevel function that will set a finalizer or if we restructure code so we will exit finalizer scope, i.e. 
```
join $ runAgentWithFinalizer .. `onException` finalize -- finally will no longer work here
```

hyperthunk:
> So we add additional layer if runAgentWithFinalizer calls runAgent. This can be resolved by introducing a single toplevel function that will set a finalizer or if we restructure code so we will exit finalizer scope, i.e.

Hmm yes okay, I'd totally missed that. I think the code needs re-structuring then. Let me go away and take a look at that. I think what we really want here is a top level `finally` only, and then masking whilst executing the finalizer. Do you agree?

qnikst:
Yes, I agree with that.

PR #318, opened 2018-11-08
relax upper bounds a little more

hyperthunk:
This is required to get downstream libraries (such as distributed-process-{task, execution, fsm} etc) to build in travis with a new-ish stack release (e.g. lts-11.13). Without it we hit errors such as the following:

```
Error: While constructing the build
plan, the following exceptions were
encountered:
In the dependencies
for distributed-process-0.7.4:
    containers-0.6.0.1 from stack
                       configuration
                       does not
                       match >=0.5 && <0.6 
                       (latest matching
                       version
                       is 0.5.11.0)
    stm-2.5.0.0 from stack configuration
                does not
                match >=2.4 && <2.5 
                (latest matching version
                is 2.4.5.1)
needed due
to distributed-process-fsm-0.0.1
    -> distributed-process-0.7.4
```

### DIFF
--- a/distributed-process.cabal
+++ b/distributed-process.cabal
@@ -45,7 +45,7 @@ Library
                      binary >= 0.6.3 && < 0.10,
                      hashable >= 1.2.0.5 && < 1.3,
                      network-transport >= 0.4.1.0 && < 0.6,
-                     stm >= 2.4 && < 2.5,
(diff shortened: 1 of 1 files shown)


hyperthunk:
I'm assuming this will need to be done in distributed-static too...?

PR #319, opened 2018-11-10
make it possible to verify against latest snapshot on stackage 

hyperthunk:
Should be able to run successfully with 

```
stack build --stack-yaml stack-nightly.yaml --resolver nightly --haddock --test --bench --no-run-benchmarks 
```

Now, please note that there are compilation failures here, and I've submitted it like this deliberately so...

We seem to be bumping into GHC being unhappy about pattern matches that can fail, and have no `MonadFail` instance in the transformer stack for `Process`. I have no idea how to add it, whether adding it is a good thing, or what the alternative(s) look(s) like...

So please advise on this so I can tidy this PR. Notably I've fudged the test cases for some compilation failures, but I'm not sure what the right thing to do is here so want further advice before continuing...

Cheers,

### DIFF
--- a/src/Control/Distributed/Process/Internal/Primitives.hs
+++ b/src/Control/Distributed/Process/Internal/Primitives.hs
@@ -403,7 +403,8 @@ receiveWait ms = do
   mProc <- liftIO $ dequeue queue Blocking (map unMatch ms)
   case mProc of
     Just proc' -> proc'
-    Nothing    -> error "Well... That wasn't supposed to happen! o_O"
+    Nothing    -> die $ "System Invariant Violation: CQueue.hs returned `Nothing` "
+                     ++ "in the absence of a timeout value."
 
 -- | Like 'receiveWait' but with a timeout.
 --
--- a/src/Control/Distributed/Process/Internal/Primitives.hs
+++ b/src/Control/Distributed/Process/Internal/Primitives.hs
@@ -400,8 +400,10 @@ newtype Match b = Match { unMatch :: MatchOn Message (Process b) }
 receiveWait :: [Match b] -> Process b
 receiveWait ms = do
   queue <- processQueue <$> ask
-  Just proc <- liftIO $ dequeue queue Blocking (map unMatch ms)
-  proc
+  mProc <- liftIO $ dequeue queue Blocking (map unMatch ms)
+  case mProc of
+    Just proc' -> proc'
+    Nothing    -> error "Well... That wasn't supposed to happen! o_O"
 
 -- | Like 'receiveWait' but with a timeout.
 --
--- a/src/Control/Distributed/Process/Management/Internal/Trace/Primitives.hs
+++ b/src/Control/Distributed/Process/Management/Internal/Trace/Primitives.hs
@@ -33,6 +33,7 @@ import Control.Distributed.Process.Internal.Primitives
   ( whereis
   , newChan
   , receiveChan
+  , die
   )
 import Control.Distributed.Process.Management.Internal.Trace.Types
   ( TraceArg(..)
@@ -168,6 +169,10 @@ withRegisteredTracer act = do
   withLocalTracer $ \t -> liftIO $ Tracer.getCurrentTraceClient t sp
   currentTracer <- receiveChan rp
   case currentTracer of
-    Nothing  -> do { (Just p') <- whereis "tracer.initial"; act p' }
+    Nothing  -> do mTP <- whereis "tracer.initial"
+                   -- NB: this should NOT ever happen, but forcing pattern matches
+                   --     is not considered cool in later versions of MonadFail
+                   case mTP of
+                     Just p' -> act p'
+                     Nothing -> die "Initial Tracer Name Not Registered"
     (Just p) -> act p
-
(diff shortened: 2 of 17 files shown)


hyperthunk:
Please also note that according to https://travis-ci.org/haskell-distributed/distributed-process/jobs/453317742, there are other compilation errors too (this one due to name shadowing...)

hyperthunk:
Background information - the stackage documentation suggests that running this command is a useful way to verify whether or not your library/executable program works with the latest of everything. I would like to get all the Cloud Haskell ecosystem up to date viz dependencies against the latest base and other libraries, and relax our upper bounds across all projects consistently (I am thinking we should either include this check against the latest nightlies in all distributed-process-X project travis configurations).

src/Control/Distributed/Process/Internal/Primitives.hs
```diff
@@ -400,8 +400,10 @@ newtype Match b = Match { unMatch :: MatchOn Message (Process b) }
 receiveWait :: [Match b] -> Process b
 receiveWait ms = do
   queue <- processQueue <$> ask
-  Just proc <- liftIO $ dequeue queue Blocking (map unMatch ms)
-  proc
+  mProc <- liftIO $ dequeue queue Blocking (map unMatch ms)
+  case mProc of
+    Just proc' -> proc'
+    Nothing    -> error "Well... That wasn't supposed to happen! o_O"
```
hyperthunk: @qnikst - clearly this isn't the right thing to do here! 🥇 

So what I can do is call the tracer here and debug log the fact that we saw an invariant in the cqueue code fail, and then call die with the appropriate message. That will kill the calling process in the correct way (with an exit signal). 

I'll cook that up and resubmit...

src/Control/Distributed/Process/Management/Internal/Trace/Primitives.hs
```diff
@@ -168,6 +169,10 @@ withRegisteredTracer act = do
   withLocalTracer $ \t -> liftIO $ Tracer.getCurrentTraceClient t sp
   currentTracer <- receiveChan rp
   case currentTracer of
-    Nothing  -> do { (Just p') <- whereis "tracer.initial"; act p' }
+    Nothing  -> do mTP <- whereis "tracer.initial"
+                   -- NB: this should NOT ever happen, but forcing pattern matches
+                   --     is not considered cool in later versions of MonadFail
+                   case mTP of
+                     Just p' -> act p'
+                     Nothing -> die "Initial Tracer Name Not Registered"
```
qnikst: can we add something like: `This should never happen` or `invariant violation` to the message. It would be nice if a user will get enough info to understand what had happened in this case.

src/Control/Distributed/Process/Internal/Primitives.hs
qnikst: It's better to elaborate on this error a bit more so if for some reason (that may appear in the future this error will be shown, a user will not be surprised.
In general, it makes sense to have an error or better `die` here, because a type of the dequeue is not precise enough to tell us that this case is not possible.

src/Control/Distributed/Process/Management/Internal/Trace/Primitives.hs
```diff
@@ -168,6 +169,10 @@ withRegisteredTracer act = do
   withLocalTracer $ \t -> liftIO $ Tracer.getCurrentTraceClient t sp
   currentTracer <- receiveChan rp
   case currentTracer of
-    Nothing  -> do { (Just p') <- whereis "tracer.initial"; act p' }
+    Nothing  -> do mTP <- whereis "tracer.initial"
```
hyperthunk: Hmmm. Although this state invariant shouldn't ever happen, I don't think we should die here. The `Debug.hs` module depends on this code, as do a lot of other modules indirectly. If there is no initial tracer, that means that one of three things has happened...

1. the initial tracer started by `startServiceProcesses` was killed
2. an exception killed `startDefaultTracer` before we could register the name
3. someone unregistered the name

Any of those cases represents a systemic problem, but fundamentally it's not an issue we should solve here. We can open a separate issue to discuss whether the node controller should monitor and restart service processes.

hyperthunk:
@qnikst - since moving to the latest GHC and mtl and so on, means that we no longer have `MonadFail` in our transformer stack, I'm assuming that the second major version will need a bump, to 0.8.0, yes?

hyperthunk:
I will need to use preprocessor blocks once a newer version of network-transport-tcp is released, since I'm guessing we do want to support older versions...?

qnikst:
> I will need to use preprocessor blocks once a newer version of network-transport-tcp is released, since I'm guessing we do want to support older versions...?

if newer network-transport supports all required GHC versions then I think it's safe to drop support of the other one, unless it's explicitly requested by the users. Supporting multiple GHC versions makes total sense, but supporting many versions of the dependencies doesn't. As far as I know no package supports LTS and backporting of the patches, so actually the only way to not keep old bugs is to use latest version of  the package.

hyperthunk:
> 
> if newer network-transport supports all required GHC versions then I think it's safe to drop support of the other one, unless it's explicitly requested by the users. Supporting multiple GHC versions makes total sense, but supporting many versions of the dependencies doesn't. As far as I know no package supports LTS and backporting of the patches, so actually the only way to not keep old bugs is to use latest version of the package.

So for older versions, would we stick to using the -location element in the stack configurations, because the newer n-t-tcp won't be in those lts releases?

qnikst:
Oh.. and what prevents n-t-tcp from being there? If that is not solvable, then supporting multiple versions is fine.

hyperthunk:
So... we have a pass from CI...

Over to you for review of the changes, @qnikst

PR #320, opened 2018-11-11
Improvements to build and test

hyperthunk:
This set of patches should

* reduce compilation warnings
* remove use of deprecated APIs in our test suites
* make travis test the last four major GHC versions
* declare the tested GHC versions correctly in the cabal file


hyperthunk:
I'd also suggest that once we merge #319, that we update the version-less stack.yaml to test against nightly-2018-11-11. In which case, we should probably look at this PR after that one.

hyperthunk:
Let's merge with the latest upstream changes...

Issue #241, opened 2015-06-22
Document guarantees provided by Management API


qnikst:
It worth documenting what are the guarantees that management API provides:
1. could messages be lost
2. could messages be reordered
3. and more complicated: is there any ordering/reliability guarantees between different Agents.

hyperthunk:
There are no ordering guarantees between Agents, no. The event bus is serial and single threaded, so we should have *some* guarantees, but not many. Anything that is published by the NC will be treated in FIFO order. There are no ordering guarantees pertaining to entries published to the event bus by other processes or agents though.

Can messages be lost? Locally, no, I don't think so. Unless the forkIO thread on which the event bus relies is killed, in which case we've got bigger problems anyway. Perhaps all the event state should be isolated further? This is similar to the problem of "what happens if the NC thread terminates?" I think.

Could messages be reordered: No, I don't see how they could.

hyperthunk:
@qnikst - how do you want this documented, and where? I'd like to close this down so that we can move on with blocked PRs...

hyperthunk:
@qnikst ... ping! I'd like to know where you want these things documented, and I'll provide a PR today or tomorrow. I'd like to get on with #265 and #184 so we can look at a minor/patch release, so I can stabilise the version of CH on which various other libraries rely (i.e., to resolve the discussion in https://github.com/haskell-distributed/distributed-process-extras/pull/6#issuecomment-191446231).

More importantly, do you agree with all of the following assertions about this....

1. since the event bus uses STM broadcast channels to communicate with agents, no message written to the bus successfully can be lost
2. agents can also receive messages via their mailboxes - these are subject to the same guarantees as all inter-process message sending
3. messages dispatched on an STM broadcast channel (i.e., management event bus) are guaranteed to be delivered with the same FIFO ordering guarantees that exist between two communicating processes, such that communication from the node controller threads (i.e., MxEvent's) will never be re-ordered, but messages dispatched to the event bus by other processes (including, but not limited to agents) are only guaranteed to be ordered between one sender and one receiver.
4. no guarantee exists for the ordering in which messages sent to an agent's mailbox will be delivered, vs messages dispatched via the event bus

I can point to the code fragments which I believe back this up, if that's useful...

qnikst:
I think I agree with all of the points, and current implementation match those.

hyperthunk:
Okay awesome, I'll add those points to the docs for Debug/Management and submit a PR.

PR #322, opened 2018-11-12
Prevent registered names from leaking

hyperthunk:
This patch subsumes #184, and incorporates #321 (which I have submitted separately, since it's less contentious). 

I have taken the code @qnikst wrote for #184, but refactored it to use the label tracking approach suggested by @mboes, which is not only simpler, but the Map handling code in the tweag:DP-100 branch is incorrect. I have also kept the agent code inside of the Node.hs module, as suggested elsewhere in the review for that PR.

This patch also incorporates the [tests](https://github.com/haskell-distributed/distributed-process-tests/pull/15) written by @qnikst (before the repos were reorganised), and I've included some additional tests in the Mx.hs test suite, which verifies that the ordering constraints are observed.

Although I have not managed to close out #241 yet, there is no reason to believe that the management event bus does not hold to the following assertion (made in the comments for that issue/ticket):

> messages dispatched on an STM broadcast channel (i.e., management event bus) are guaranteed to be delivered with the same FIFO ordering guarantees that exist between two communicating processes, such that communication from the node controller threads (i.e., MxEvent's) will never be re-ordered, but messages dispatched to the event bus by other processes (including, but not limited to agents) are only guaranteed to be ordered between one sender and one receiver.

### DIFF
--- a/distributed-process-tests/distributed-process-tests.cabal
+++ b/distributed-process-tests/distributed-process-tests.cabal
@@ -120,7 +120,7 @@ Test-Suite TestMxInMemory
   CPP-Options:       -DTEST_SUITE_MODULE=Control.Distributed.Process.Tests.Mx
   Build-Depends:     base >= 4.4 && < 5,
                      distributed-process-tests,
-                     network >= 2.3 && < 2.7,
+                     network >= 2.3 && < 2.9,
                      network-transport >= 0.4.1.0 && < 0.6,
                      network-transport-inmemory >= 0.5,
                      test-framework >= 0.6 && < 0.9
@@ -134,7 +134,7 @@ Test-Suite TestTracingInMemory
   CPP-Options:       -DTEST_SUITE_MODULE=Control.Distributed.Process.Tests.Tracing
   Build-Depends:     base >= 4.4 && < 5,
                      distributed-process-tests,
-                     network >= 2.3 && < 2.7,
+                     network >= 2.3 && < 2.9,
                      network-transport >= 0.4.1.0 && < 0.6,
                      network-transport-inmemory >= 0.5,
                      test-framework >= 0.6 && < 0.9
--- a/distributed-process-tests/distributed-process-tests.cabal
+++ b/distributed-process-tests/distributed-process-tests.cabal
@@ -155,17 +155,3 @@ Test-Suite TestMxInTCP
   Extensions:        CPP
   ghc-options:       -Wall -threaded -rtsopts -with-rtsopts=-N -fno-warn-unused-do-bind
   HS-Source-Dirs:    tests
-
-Test-Suite TestTracingInTCP
-  Type:              exitcode-stdio-1.0
-  Main-Is:           runInMemory.hs
-  CPP-Options:       -DTEST_SUITE_MODULE=Control.Distributed.Process.Tests.Tracing
-  Build-Depends:     base >= 4.4 && < 5,
-                     distributed-process-tests,
-                     network >= 2.3 && < 2.9,
-                     network-transport >= 0.4.1.0 && < 0.6,
-                     network-transport-inmemory >= 0.5,
-                     test-framework >= 0.6 && < 0.9
-  Extensions:        CPP
-  ghc-options:       -Wall -eventlog -threaded -rtsopts -with-rtsopts=-N -fno-warn-unused-do-bind
-  HS-Source-Dirs:    tests
--- a/distributed-process-tests/distributed-process-tests.cabal
+++ b/distributed-process-tests/distributed-process-tests.cabal
@@ -114,8 +114,7 @@ Test-Suite TestStats
   ghc-options:       -Wall -eventlog -threaded -rtsopts -with-rtsopts=-N -fno-warn-unused-do-bind
   HS-Source-Dirs:    tests
 
-
-Test-Suite TestMx
+Test-Suite TestMxInMemory
   Type:              exitcode-stdio-1.0
   Main-Is:           runInMemory.hs
   CPP-Options:       -DTEST_SUITE_MODULE=Control.Distributed.Process.Tests.Mx
@@ -129,7 +128,35 @@ Test-Suite TestMx
   ghc-options:       -Wall -threaded -rtsopts -with-rtsopts=-N -fno-warn-unused-do-bind
   HS-Source-Dirs:    tests
 
-Test-Suite TestTracing
+Test-Suite TestTracingInMemory
+  Type:              exitcode-stdio-1.0
+  Main-Is:           runInMemory.hs
+  CPP-Options:       -DTEST_SUITE_MODULE=Control.Distributed.Process.Tests.Tracing
+  Build-Depends:     base >= 4.4 && < 5,
+                     distributed-process-tests,
+                     network >= 2.3 && < 2.7,
+                     network-transport >= 0.4.1.0 && < 0.6,
+                     network-transport-inmemory >= 0.5,
+                     test-framework >= 0.6 && < 0.9
+  Extensions:        CPP
+  ghc-options:       -Wall -eventlog -threaded -rtsopts -with-rtsopts=-N -fno-warn-unused-do-bind
+  HS-Source-Dirs:    tests
+
+Test-Suite TestMxInTCP
+  Type:              exitcode-stdio-1.0
+  Main-Is:           runInMemory.hs
+  CPP-Options:       -DTEST_SUITE_MODULE=Control.Distributed.Process.Tests.Mx
+  Build-Depends:     base >= 4.4 && < 5,
+                     distributed-process-tests,
+                     network >= 2.3 && < 2.7,
+                     network-transport >= 0.4.1.0 && < 0.6,
+                     network-transport-inmemory >= 0.5,
+                     test-framework >= 0.6 && < 0.9
+  Extensions:        CPP
+  ghc-options:       -Wall -threaded -rtsopts -with-rtsopts=-N -fno-warn-unused-do-bind
+  HS-Source-Dirs:    tests
+
+Test-Suite TestTracingInTCP
   Type:              exitcode-stdio-1.0
   Main-Is:           runInMemory.hs
   CPP-Options:       -DTEST_SUITE_MODULE=Control.Distributed.Process.Tests.Tracing
--- a/distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
+++ b/distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
@@ -1505,7 +1505,7 @@ testRegistryMonitoring TestTransport{..} = do
     -- away, and thus not be subjected to a 20 second delay. The value of 4
     -- seconds appears to work optimally on osx and across several linux distros
     -- running in virtual machines (which is essentially what we do in CI)
-    receiveTimeout 4000000 [ matchAny return ]
+    void $ receiveTimeout 4000000 [ matchAny return ]
 
   -- This delay doesn't serve much purpose in the happy path, however if some
   -- future patch breaks the cooperative behaviour of node controllers viz
--- a/distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
+++ b/distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
@@ -37,9 +37,11 @@ import Control.Distributed.Process.Internal.Types
   , createUnencodedMessage
   )
 import Control.Distributed.Process.Node
+import Control.Distributed.Process.Debug
+import Control.Distributed.Process.Management.Internal.Types
 import Control.Distributed.Process.Serializable (Serializable)
 
-import Test.HUnit (Assertion, assertFailure)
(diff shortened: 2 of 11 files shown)


src/Control/Distributed/Process/Node.hs
```diff
@@ -310,6 +325,45 @@ startDefaultTracer node' = do
...
+--
+
+registryMonitorAgent :: Process ProcessId
+registryMonitorAgent = do
+  nid <- getSelfNode
+  -- For each process the map associates the 'MonitorRef' used to monitor it and
+  -- the amount of labels associated with it.
+  mxAgent registryMonitorAgentId (Map.empty :: Map String MonitorRef)
+    [ mxSink $ \(ProcessMonitorNotification mref _ _) -> do
+        mxUpdateLocal $ Map.filter (== mref)
```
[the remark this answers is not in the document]
hyperthunk: Erm... Yes. It should. Lemme fix that. 👍

hyperthunk:
Hold tight... Fix to the broken test(s) coming...

hyperthunk:
Hoo-bloomin-rah! Looks like we've finally got some stability in the tests...

@qnikst, whenever you've got time, can you take a look at the updated patch set please? Also, we should probably squash these commits down, as the PR is a right mess of interventions now...

distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
```diff
@@ -1369,6 +1372,116 @@ testExitRemote TestTransport{..} = do
   takeMVar supervisedDone
   takeMVar supervisorDone
 
+testRegistryMonitoring :: TestTransport -> Assertion
+testRegistryMonitoring TestTransport{..} = do
+  localNode <- newLocalNode testTransport initRemoteTable
+  remoteNode <- newLocalNode testTransport initRemoteTable
+  return ()
```
qnikst: [nitpick] ?

distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
```diff
@@ -1369,6 +1372,116 @@ testExitRemote TestTransport{..} = do
...
+
+  -- Local process. Test if local process will be removed from
+  -- registry when it dies.
+  box <- newEmptyMVar
+  runProcess localNode $ do
+    pid <- spawnLocal $ do
```
qnikst: [nitpick] redundant `do`

distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
```diff
@@ -1369,6 +1372,116 @@ testExitRemote TestTransport{..} = do
...
+        w1 <- whereis "test-3" :: Process (Maybe ProcessId)
+        w2 <- whereis "test-4" :: Process (Maybe ProcessId)
+        forM_ (w1 <|> w2) (const waitpoll)
+    in do register "test-3" remote2
+          register "test-4" remote2
+          send remote2 ()
+          waitpoll
+          return ()
+
+{- XXX: waiting including patch for nsend for remote process
```
qnikst: I think that the patch has landed, if not, can we add a link to that?

distributed-process-tests/src/Control/Distributed/Process/Tests/Mx.hs
```diff
@@ -174,32 +199,159 @@ testAgentEventHandling result = do
...
+testMxRegMon :: LocalNode -> TestResult () -> Process ()
+testMxRegMon remoteNode result = do
+  ensure (stash result ()) $ do
+    -- ensure that when a registered process dies, we get a notification that
+    -- it has been unregistered as well as seeing the name get removed
+    let label1 = "aaaaa"
+    let label2 = "bbbbb"
+    let isValid l = l == label1 || l == label2
+    let agentLabel = "mxRegMon-agent"
+    let delay = 1000000
```
qnikst: it's a pity that we have to use `delay` but ok for now.

distributed-process-tests/distributed-process-tests.cabal
```diff
@@ -114,8 +114,35 @@ Test-Suite TestStats
...
+                     network >= 2.3 && < 2.7,
+                     network-transport >= 0.4.1.0 && < 0.6,
+                     network-transport-inmemory >= 0.5,
+                     test-framework >= 0.6 && < 0.9
+  Extensions:        CPP
+  ghc-options:       -Wall -eventlog -threaded -rtsopts -with-rtsopts=-N -fno-warn-unused-do-bind
+  HS-Source-Dirs:    tests
 
-Test-Suite TestMx
+Test-Suite TestMxInTCP
```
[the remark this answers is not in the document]
hyperthunk: Agreed, I will adjust that this afternoon

distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
hyperthunk: About...? 😂

distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
hyperthunk: Okay I'll lint over the changes. Although the linter is generally unhappy about *lots* of things, so maybe just this function and then a separate cosmetic PR...

distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
hyperthunk: I thought so too, just wanted confirmation. I'll uncomment and we'll see what Travis thinks!

distributed-process-tests/src/Control/Distributed/Process/Tests/Mx.hs
hyperthunk: Well, the real issue is that we can see events related to monitoring agents from other monitoring agents...

So one solution is to run each test case in a new local node. That might actually be a better approach here, tbh. Maybe I will re-cook the patch accordingly and resubmit... Or do that as part of the next patch I'm working on, which has management/tracing fixes and improvements?

distributed-process-tests/distributed-process-tests.cabal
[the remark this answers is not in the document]
hyperthunk: Though the CH test case says `InTCP`, so I chose this for consistency. Perhaps we could change both in a separate patch...?

distributed-process-tests/src/Control/Distributed/Process/Tests/Mx.hs
qnikst: It's ok to have that in another PR.

distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
qnikst: ops.. is `return ()` really needed there?

hyperthunk: No idea, it's in the original tests, which I'm pretty sure you wrote! :D

I don't think it's actually needed, no, so I'll go remove it and we'll see if the tests remain stable. They should, I mean, we're not exactly deterministic about giving up time slices so this expression shouldn't make any difference.

qnikst: Hah.. maybe it makes use of the tracing and overrived `>>=`. Ok lets keep that.

hyperthunk:
```patch
diff --git a/distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs b/distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
index 91e1323..7430562 100644
--- a/distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
+++ b/distributed-process-tests/src/Control/Distributed/Process/Tests/CH.hs
@@ -1476,6 +1476,37 @@ testExitRemote TestTransport{..} = do
   takeMVar supervisedDone
   takeMVar supervisorDone
 
+testRegisterRemote :: TestTransport -> Assertion
+testRegisterRemote TestTransport{..} = do
+  node1 <- newLocalNode testTransport initRemoteTable
+  node2 <- newLocalNode testTransport initRemoteTable
+  done <- newEmptyMVar
+
+  pid <- forkProcess node1 $ getSelfPid >>= runUntilRegistered
+
+  runProcess node2 $ do
+    register regName pid
+    liftIO $ putMVar done ()
+
+  takeMVar done
+  threadDelay 400000  -- TODO: this is racy, but we want to avoid using mx for now...
+
+  regHere <- newEmptyMVar
+
+  runProcess node2 $ whereis regName >>= liftIO . putMVar regHere
+  res <- takeMVar regHere
+  assertBool "expected Nothing, but process id was still registered" (res == Nothing)
+
+  where
+    runUntilRegistered us = do
+      mPid <- whereis regName
+      case mPid of
+        Nothing -> runUntilRegistered us
+        Just p
+          | p == us -> return ()
+
+    regName = "testRegisterRemote"
+
 testUnsafeSend :: TestTransport -> Assertion
 testUnsafeSend TestTransport{..} = do
   serverAddr <- newEmptyMVar
@@ -1677,6 +1708,7 @@ tests testtrans = return [
       , testCase "MaskRestoreScope"    (testMaskRestoreScope    testtrans)
       , testCase "ExitLocal"           (testExitLocal           testtrans)
       , testCase "ExitRemote"          (testExitRemote          testtrans)
+      , testCase "RegisterDiedRemote"  (testRegisterRemote      testtrans)
       , testCase "TextCallLocal"       (testCallLocal           testtrans)
       -- Unsafe Primitives
       , testCase "TestUnsafeSend"      (testUnsafeSend          testtrans)
```

Right... This patch (against master) seems to fail reliably. Let's incorporate that into this PR and see if I can demonstrate that the issue is resolved. I'm also going to see what happens when we drop the more complex test case into master.

hyperthunk:
> Right... This patch (against master) seems to fail reliably. Let's incorporate that into this PR and see if I can demonstrate that the issue is resolved. I'm also going to see what happens when we drop the more complex test case into master.

Indeed, the original test cases still pass when I don't start the mx agent in the node controller, whereas the simpler test case fails reliably. I'm going to remove the more complex test case, though we can roll it back in if you have suggestions about how to make it work.

hyperthunk:
Oddly the monitoring doesn't fix the failing test case. Tracing shows that the map is updated correctly, so I'm going to trace the `whereIs` handling in the NC to see if there's a race in the test case...

hyperthunk:
>  I'm going to comment out updateRemote in ncEffectRegister and see which test case, if either, breaks...

Oh yes, they do indeed.

qnikst:
@hyperthunk what you wrote makes sense. If I recall correctly there were some problems at the time when initial PR was written, but maybe there were some other issues that prevented the logic from working correctly fixed after that.

hyperthunk:
> @hyperthunk what you wrote makes sense. If I recall correctly there were some problems at the time when initial PR was written, but maybe there were some other issues that prevented the logic from working correctly fixed after that.

Possibly, yes. I can leave the monitoring in, as it certainly won't do any harm. But it is not essential to this patch set. You can comment out the `void $ registryMonitorAgent` expression in `startServiceProcesses` and all the test cases pass.

I now have what I believe is a viable patch anyway. I will push in a moment and we can review the final product.

hyperthunk:
Okay, so assuming CI is happy, I think we're good to go. We have fixed the tracing of `MxUnregistered` in the case where a (local or remote) process that is registered dies, and we have a test case for that in the `Mx.hs` suite. We have a new test case verifying that the node controllers collaborate to inform us if a remote process registered on our local node dies, and this test case fails in (what I think is) a reasonable way if we deliberately break the code.

There are also some generic fixes to the `Mx.hs` test suite to remove potential races and name clashes, and a few cosmetic entries. Finally, this PR also runs the Mx and Tracing test suites with both transports.

Finally, I have rolled back the move to run the `Tracing.hs` using the TCP transport, since in master we only use the in-memory one, and I'm getting intermittent failures in one test case, both locally and in CI. I have raised #323 to address this separately.

@qnikst - since I've made quite a few changes, would you mind doing the merge once you're happy this all makes sense? I'm probably tree blind by now, having worked on it.

facundominguez:
How does the current code ensure that when there is a connection failure, the label pointing to the remote process is unregistered?

facundominguez:
Hm, spent some time looking at the code and it looks ok.

> How does the code before this patch ensure that when there is a connection failure, the label pointing to the remote process is unregistered?
>
> Is ncEffectDied going to be called if no process in the local node is monitoring the remote node?

The NC produces a `Died` message to itself when the transport detects an error. ncEffecDied removes the nodes from `registeredHereFor` and `registeredOnNodesFor`.

Answering myself, if the monitor agent is not necessary, I'd leave it out to avoid confusion. Every person reading the code would ask what it is needed for and there is no motivation to maintain it.

hyperthunk:
Agree with all that @facundominguez, I'll remove the code. We seem to have picked up an odd CI failure anyway so I'll resolve both and ping back

hyperthunk:
Right... @facundominguez @qnikst - if you're both happy, I think we can finally get this one merged!

Issue #333, opened 2018-11-24
The remote registry test case is far too racy


hyperthunk:
I'm not sure why we're seeing this happen more consistently now, but I'll investigate and fix

hyperthunk:
It seems something was really broken for these failures to occur...

The first is against ghc-8.4.4, in TestCHInTCP, and [here is the build log](https://travis-ci.org/haskell-distributed/distributed-process/jobs/458999359). Note:
```
  RegistryMonitoring: [Failed]
expected (Just pid)
```

Which means the following code failed:

```haskell
  runProcess node2 $ do
    register regName pid
    res <- whereis regName
    liftIO $ do
      putMVar waitH ()
      assertBool "expected (Just pid)" $ res == (Just pid)
```

I *really* cannot see how this failed without us generating a `ProcessRegistrationException` in the `register` call... :/

The next failure was a general test failure, logged [here](https://travis-ci.org/haskell-distributed/distributed-process/jobs/458999357) against ghc-7.10.3, in TestCHInMemory

```
RegistryMonitoring: [Failed]
expected Nothing, but process still registered)
```

This means we waited and saw the the name was registered, told the registered pid to exit (by writing to the mvar it was waiting on), waited four seconds, then executed a `whereIsRemoteAsync` call and waited for a `WhereIsReply` with a twenty second timeout, then ran *another* call to `whereis` (the local version, as the registration took place on this node), and we still had a process registered in our local registry...

What I'm reading from that second failure, is that either (a) despite waiting for 20 seconds to see the right update, the network transport layer had not notified our node yet that the remote process has died, or (b) our cleanup code isn't covering some odd edge case.

hyperthunk:
Sadly this test failures are not something I've seen before (since the test code was added), cannot be reproduced locally - I've been running the test suite in a loop waiting for failures under OSX High Sierra on an Intel Core i5 (dual core) machine, and on an identical machine running docker, using the same Ubuntu Trusty image that travis uses.  So far 7 hours of continuous testing on both platforms and I've not been able to reproduce...

hyperthunk:
Unfortunately, adding a monitor to the test case (to ensure the remote pid is down) would break the invariants upon which the test case is predicated, since we're trying to assert that we handle the deaths of remote processes that are registered locally without the requirement for explicit monitoring.

[...]

PR #335, opened 2018-11-26
Rework some tests to use a genuinely remote node

hyperthunk:
I _think_ the approach outlined in this patch set is a reasonable solution for #333. I would envisage moving a few of the more racy/flaky test cases over to this approach. I'd like some feedback on the general advisability if possible, before moving a few other test cases and merging.

It would be worth seeing if my argument in #333 about the scheduler creating the failure conditions for our test case, holds up to scrutiny. 

I would prefer if this were ultimately in some kind of integration-test package, executed on a build pipeline once the regular test suites are complete. Sadly travis doesn't really have a proper notion of pipelines (frankly, apart from Jenkins/Hudson and Bamboo, I've seen for CI systems that offer a strong build pipeline capability, and none for free)...

And before someone says it, I *absolutely* think we should start doing model based testing in addition to this approach. I've been looking at https://github.com/advancedtelematic/quickcheck-state-machine-distributed, and considering some other options.

Anyway... Does this patch make sense, and do you agree with my thoughts in #333? Because if you think I'm barking up the wrong tree, then this is a pointless additional of technical debt, and complicates the CI build quite a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,8 @@ script:
   export TEST_PACKAGE="distributed-process-tests:"
   export DISTRIBUTED_PROCESS_TEST_NODE="1"
 - stack $ARGS build --no-terminal ${TEST_PACKAGE}RemoteHost --flag ${TEST_PACKAGE}remote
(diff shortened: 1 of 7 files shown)


hyperthunk:
Putting this on hold for the time being...

PR #364, opened 2021-11-08
Squash warnings in GHC 8.10.7, switch Serializable to type synonym.

davidsd:


joelmccracken:
I wish i knew if there'd be any consequences to switching serializable to a type instance

davidsd:
This is a good question. The construction that was there is a kind of "newtype" for a class. I believe it can be used in conjunction with various `Dict` tricks, see for example [here](https://www.reddit.com/r/haskell/comments/mv1zx2/implementing_reifiesgiven_for_class_with/). However, I am kind of doubtful that users of Cloud Haskell are doing these things with `Serializable`. In my experience, a type synonym and `ConstraintKinds` is simpler, results in good type inference, and doesn't yield a bunch of GHC warnings.

It would be nice to have a better concrete understanding of what users are doing with Cloud Haskell. There aren't a huge number of reverse dependencies for `distributed-process` and aren't part of the `distributed-process-` family. [Here's a list of them](https://packdeps.haskellers.com/reverse/distributed-process). Not sure what to do with this.

joelmccracken:
i'm leaning toward just mearging this and making the next big release a breaking change. WDYT?

davidsd:
I think that makes sense.

davidsd:
By now, other people have submitted pull requests to move to later GHC versions, and master has moved on. I still don't know what the effects of switching Serializable to a type synonym might be, so I'm going to just close this. Maybe I'll get a chance to ask an expert at some point.

PR #365, opened 2021-11-09
Get things working with ghc 9.0.1

davidsd:
Before pushing this to hackage, we should probably update the network-transport libraries and remove the rematch dependency, so that the extra-deps in stack.yaml can be removed. We should also wait for an LTS with GHC 9.0.2. Pushing this now to check that CI is happy.


davidsd:
I am not sure what the effects of turning Serializable into a type synonym are, so I am closing this.

PR #367, opened 2022-10-11
Fixed compilation error on GHC 9.0.2

matt2718:
I've corrected the Template Haskell code to account for the new behavior of TyVarBndr (with CPP macros, so it should be backwards-compatible). distributed-process now builds with the latest LTS version of stack.

All unit tests have passed locally.

### DIFF
--- a/.github/workflows/distributed-process-ci.yml
+++ b/.github/workflows/distributed-process-ci.yml
@@ -6,12 +6,12 @@ jobs:
     strategy:
       matrix:
         ghcVersion:
-          [ "7.10.3"
-          , "8.2.2"
+          [ "8.2.2"
           , "8.4.4"
           , "8.6.5"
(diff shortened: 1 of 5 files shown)


davidsd:
Can you also adjust the ci tests so that these changes can be verified by ci? See https://github.com/haskell-distributed/distributed-process/pull/365 for an example. (That pull request implements a lot of the same things you did, but it also makes a bunch more changes, so I'd be more comfortable just merging yours.)

davidsd:
I think you can safely remove 7.10.3 from ci as well.

matt2718:
CI results: https://github.com/matt2718/distributed-process/actions/runs/3231284906
(idk why they're not running in the request thread)

davidsd:
Thanks, @matt2718 !

PR #371, opened 2023-05-01
Got tests running locally for ghc 9.4.5

davidsd:


davidsd:
The stack-ghc-9.4.5.yaml file uses a nightly resolver, instead of lts. But given how rarely we make these updates, we might as well go ahead and support 9.4.5 now.

PR #372, opened 2023-05-02
Warnings 9.4.5

davidsd:
This branch fixes a bunch of compilation warnings. The only nontrivial change is that `Serializable` is now a type synonym instead of a separate class. This could potentially affect users who try to use `Serializable` in a partially applied form -- they would have to define their own class.


hyperthunk:
I am fine with this impact as we do not have widespread users to be affected. @facundominguez what are your thoughts?

facundominguez:
Looks good to me. Maybe it is worth checking if it affects manipulating `Dict (Serializable a)` values that may appear when using static pointers. I'll be away from a pc until the 19th.

davidsd:
One can definitely still manipulate `Dict (Serializable a)` if `Serializable a` is a type synonym. If someone still runs into an issue, they could of course still define their own class "newtype". I am inclined to go ahead with this change. Given that `distributed-process` has not been buildable with recent GHC's for a while, I'm guessing there won't be many users clamoring to change it back.

facundominguez (approved):

PR #374, opened 2023-06-27
Add nix-shell and hie.yaml for development purposes

JonathanLorimer:
The nix shell doesn't affect any of the cabal / stack files so I hope it isn't too intrusive. I am happy to maintain it, although it shouldn't require much maintenance. The nix shell also provides HLS, so I included an `hie.yaml` file for multi-cradle support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 dist/
+dist-newstyle/
 .cabal-sandbox
 cabal.sandbox.config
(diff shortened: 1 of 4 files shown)


davidsd:
Thanks for these contributions! I am happy to include them. I don't have any experience with nix or HLS, so I appreciate your offer to maintain these files.

PR #376, opened 2024-03-19
Update upper bound on `hashable`

LaurentRDC:
Hello,

This is a simple pull request to update the upper bound on `hashable` to include 1.4.3.0, which allows to build `distributed-process` with GHC 9.8

Cheers,
Laurent


davidsd:
Thanks!

PR #417, opened 2024-09-03
Fix disconnect remote endpoints on close endpoint

LaurentRDC:
This is taken from https://github.com/haskell-distributed/network-transport-inmemory/pull/24, but the development of network-transport-inmemory has moved here


LaurentRDC:
Fixes #418

PR #466, opened 2025-01-08
Port test suites to `tasty` (from `test-framework`)

### LINKED ISSUE #465: Migrate test suites from test-framework to tasty
The test framework used for all packages in this repository, `test-framework`, is [effectively on life support](https://github.com/haskell/test-framework/pull/73). Test suites should be ported to use `tasty` instead.

LaurentRDC:
`test-framework` is essentially on life support. We are easing maintenance burden by porting test suites to `tasty`.

Fixes #465


LaurentRDC:
In the process of touching almost every test suite, I have noticed that a lot of packages were listed as dependencies, but not used (using the `-Wunused-packages` warning). I have therefore activated this flag by default for every build target, in an effort to minimize the number of unused dependencies.

PR #467, opened 2025-02-05
General tidy up of dependencies and warnings

tmcgilchrist:
Thank you for picking up maintenance of this package.

I've done some hopefully useful (and non-controversial) fixes to cabal files, source files and CI setup.
I don't think this introduces any regressions but I have seen intermittent failures on the testsuite that are fixable by re-running the tests. The test suite needs work to make it more reliable, which I've not done here.

* a1071e6ecc03ec619b9dc5b497a31dd6c5e7e288 removes unnecessary package depdendencies.
* b601ed10a33577ff9a299b548eb213b786bb547e changes GH CI so it will only build once per PR. Previously it was running two identical builds for PRs.
* e3196db70e16e54393c17c927a6f87c26f678fe2 removes unnecessary imports, exports and variable bindings.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -2,10 +2,12 @@ name: Continuous integration
 
 on:
   push:
+    branches: ['master']
     paths-ignore:
       - 'README.md'
       - 'CONTRIBUTING.md'
   pull_request:
+    branches: ['master']
 
 jobs:
   continuous-integration:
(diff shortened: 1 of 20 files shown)


LaurentRDC:
This is great! Thank you so much

LaurentRDC:
I'm preparing releases for all packages with unreleased changes

tmcgilchrist:
Thank you @LaurentRDC

PR #488, opened 2026-09-18
Fixed an issue where `mkClosureVal` and similar functions could not be used due to overlapping instances

LaurentRDC:


LaurentRDC:
I don't understand why this isn't working for GHC 8.10 - 9.2, so I'm going to leave this be

PR #490, opened 2026-09-21
Increase throughput by arming timeouts only if necessary

LaurentRDC:
This PR increases throughput when using `receiveTimeout`, `expectedTimeout`, and `receiveChanTimeout`. In case a message is already in the process mailbox, no timer is armed. 
This can result in a 9x throughput increase for small messages, under load. Benchmarks have been added as well.

Note that the function docstrings already mentioned that this was being done.


LaurentRDC:
This is a big enough improvement that I should cut a 0.7.9 release soon

